PR #19, opened 2024-12-08
fix: update funding

castrojo:
Add yourself in here. 😄


nicknamenamenick:
I think this should be closed unless a maintainer wants to put themself if they have Github sponsors.

PR #22, opened 2024-12-10
refactor: remove bluefin references and update descriptions in workflows

RealVishy:
remove references to bluefin in the workflows, there's also some in the [changelogs.py](https://github.com/ublue-os/aurora/blob/main/.github/changelogs.py) but i didn't want to mess with it.

Let me know if i missed anything


nicknamenamenick (approved):

RealVishy:
my bad, i didn't realise that second commit would go to this pr as well

NiHaiden (approved):

Issue #28, opened 2024-12-11
Port Sunshine for remote desktop use


linkion:
Hello, I'm trying to setup sunshine to be used alongside tailscale with Aurora, but it seems that I can't install it without the workaround that's been applied on the bazzite side, because attempting to apply 
`sudo setcap cap_sys_admin+p $(readlink -f $(which sunshine))`
results in
`Failed to set capabilities on file '/usr/bin/sunshine-v2024.1208.41026': Read-only file system`

To get around this, bazzite has a system-level [systemd service](https://github.com/ublue-os/bazzite/blob/main/system_files/desktop/shared/usr/lib/systemd/system/sunshine-workaround.service) to apply this.

I believe we could just copy paste the existing just from bazzite.
https://github.com/ublue-os/bazzite/blob/main/system_files/desktop/shared/usr/share/ublue-os/just/82-bazzite-sunshine.just

AgentK9:
for anyone else who comes across this - the RPM URL you want to use for rpm-ostree is currently here: https://copr.fedorainfracloud.org/coprs/lizardbyte/stable/ for sunshine

AgentK9:
This issue should probably be closed

samexner:
> This issue should probably be closed

Why? This is still a problem. `sudo setcap cap_sys_admin+p $(readlink -f $(which sunshine))` results in the same error, which in turn keeps capture from working on Wayland. This is keeping me from using Aurora and I don't want Bazzite.

inffy:
implemented

PR #29, opened 2024-12-11
feat: Port sunshine config from bazzite

linkion:
Here is a pull request for porting over sunshine to be used within Aurora.

I don't know how bazzite was able to install sunshine with 
```
rpm-ostree install --apply-live -y sunshine
```

I did instead replace it with the [copr](https://copr.fedorainfracloud.org/coprs/lizardbyte/) provided by lizardbyte. However, because the stable copr repo doesn't have a stable build of Sunshine, I'm using the beta repo, which does have a recent build.

```
dnf5 copr enable -y lizardbyte/beta
rpm-ostree install --apply-live -y Sunshine
```

I don't necessarily know how to test this though, I have tried to run with `just run aurora stable nvidia`, but it seems that systemd isn't used so I can't test if the service runs properly, and building and running an iso with `just run-iso aurora stable` fails on installation of the OS in the vm.

The issue linked with this pull request: #28


linkion:
Sunshine is currently being pulled from Lizardbyte/beta, but we probably don't want beta software on stable.

I raised an [issue](https://github.com/LizardByte/Sunshine/issues/3448) about it with them, and they said that the copr was made recently, so it should be built into the next stable release. 

Seeing that their last stable release [[v0.23.1]](https://github.com/LizardByte/Sunshine/releases/tag/v0.23.1) was last released on April 20, it might take a while for their next release.

I'll try to pull from their latest release and install the .rpm file instead.

linkion:
Lol, nvm, they only have up to Fedora 39 on their latest release, with even their pre-releases forgoing the Fedora package. Probably because they now have the copr setup, so they removed the RPM from GitHub releases

RealVishy:
> Lol, nvm, they only have up to Fedora 39 on their latest release, with even their pre-releases forgoing the Fedora package. Probably because they now have the copr setup, so they removed the RPM from GitHub releases

Rip, should we just wait for a stable release on the COPR?

ReenigneArcher:
At this point our beta is probably less buggy than v0.23.1, and we are trying to get a new release published soon, but I can't give an accurate ETA on that.

system_files/shared/usr/lib/systemd/system/sunshine-workaround.service
ReenigneArcher, line 1: Sunshine "should" have the correct capabilities.

https://github.com/LizardByte/Sunshine/blob/1543f584abf33adfa50731b2a84083c70faaa18d/packaging/linux/fedora/Sunshine.spec#L234-L235

That was the recommendation from an rpm-ostree maintainer: https://github.com/coreos/rpm-ostree/discussions/5036#discussioncomment-10291071

linkion: Can confirm, on v2024.1211.191744 sunshine properly sets cap
(I think when I installed it before, the version might have failed to build, and I couldn't pick it up)
edit: I realize I installed it right before v2024.1211.191744, glad that this works though

I'll go ahead and remove the work around.

RealVishy:
yeah i can't get the iso to build either, i think we might still have to do some cleanup for bluefin stuff. I'll try to get that done in a pr quickly.

linkion:
> yeah i can't get the iso to build either, i think we might still have to do some cleanup for bluefin stuff. I'll try to get that done in a pr quickly.
> 
> ```
> sent 48,537,467 bytes  received 3,891 bytes  97,082,716.00 bytes/sec
> total size is 48,515,593  speedup is 1.00
> + rsync -rvK /ctx/system_files/silverblue/ /
> sending incremental file list
> rsync: [sender] change_dir "/ctx/system_files/silverblue" failed: No such file or directory (2)
> 
> sent 19 bytes  received 12 bytes  62.00 bytes/sec
> total size is 0  speedup is 0.00
> rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1338) [sender=3.3.0]
> Error: building at STEP "RUN --mount=type=cache,dst=/var/cache/rpm-ostree --mount=type=bind,from=ctx,source=/,target=/ctx /ctx/build_files/shared/build-base.sh": while running runtime: exit status 23
> error: Recipe `build` failed with exit code 23
> error: Recipe `build-iso` failed with exit code 23```
> ```

Huh, I was able to build with `just build-iso aurora stable` I think successfully, since I got an iso, though I do see some kind of error at the end
```
implantisomd5 /github/workspace/aurora_build/aurora-stable.iso
Inserting md5sum into iso image...
md5 = f7bae8f631ab4483df33cb349035fd63
Inserting fragment md5sums into iso image...
fragmd5 = ec77212b6776625a2e3dfccfd2d333d2793993cdb652abc37e6ef41163c5
frags = 20
Setting supported flag to 0
chmod ugo=r /github/workspace/aurora_build/aurora-stable.iso
cd /github/workspace/aurora_build/ && sha256sum aurora-stable.iso > aurora-stable.iso-CHECKSUM
+ [[ 1001 -gt 0 ]]
+ just sudoif chown 1001:1001 -R /var/home/patrickorave/Documents/Code/aurora
ksshaskpass: Unable to parse phrase "[sudo] password for patrickorave: "
```


But when I run `just run-iso aurora stable` and begin the installation, I get this error message on installing fedora
Screenshot_20241215_181003

inffy:
> iso build is broken for me rn, this pr should address it. #36

Iso seems to build now, but if I try to run the built iso, the installer throws an error when the installation starts.

not  sure if I  do something wrong but same thing happens even with a clean local bluefin repo clone :(

nicknamenamenick:
A bit of an FYI but [Bazzite is experimenting with having Sunshine as part of the image](https://github.com/ublue-os/bazzite/commit/a6c9c7537e4f4f2b9eafc338f3956b1e7796ce38) now due to issues how it is currently setup.

RealVishy:
> A bit of an FYI but [Bazzite is experimenting with having Sunshine as part of the image](https://github.com/ublue-os/bazzite/commit/a6c9c7537e4f4f2b9eafc338f3956b1e7796ce38) now due to issues how it is currently setup.
> 
> This might be controversial but I think we're going to stick being a little more conservative and point in the direction for Bazzite for some of these specific gaming needs.

TBH I have a couple of friends who use sunshine for remote desktop since parsec hosting doesn't work on linux, I think it'd be nice to have.

linkion:
Seems like there really isn't much interest for this to be merged, and people who want Sunshine could just rpm-ostree install it without any issues now, unlike in the case before where we needed to have a systemd workaround baked in.

I'm gonna close this

PR #40, opened 2024-12-17
Update bluefin-system.just

### LINKED ISSUE #35: Rename bluefin-cli to aurora-cli
Outside of the name, it will execute the same `ujust` script.

inffy:
move from bluefin-cli to aurora-cli just command. 

Also fix some generic text that was still referencing Bluefin (like the "changing to devmode")

fixes #35

### DIFF
--- a/just/bluefin-system.just
+++ b/just/bluefin-system.just
@@ -16,16 +16,16 @@ benchmark:
     @echo 'Running a 1 minute benchmark ...'
     stress-ng --matrix 0 -t 1m --times
 
-# Configure Bluefin-CLI Terminal Experience with Brew
+# Configure Aurora-CLI Terminal Experience with Brew
 [group('System')]
-bluefin-cli:
+aurora-cli:
     @/usr/libexec/ublue-bling.sh
 
 # alias for toggle-devmode
 devmode:
     @ujust toggle-devmode
 
-# Toggle between Bluefin and the Developer Experience
+# Toggle between Aurora and the Developer Experience
 [group('System')]
 toggle-devmode:
     #!/usr/bin/env bash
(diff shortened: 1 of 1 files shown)


nicknamenamenick (approved):

inffy:
Just a small fix as the script was still mentioning to rerun the bluefin-cli command if run again

inffy:
Actually might be better to just dismiss this one and go with t he #45 which goes much deeper than this one :)

nisel11:
> Actually might be better to just dismiss this one and go with the #45 which goes much deeper than this one :)

I thought about splitting my commit into several parts for convenience and adding this commit and #41, but haven't figured out how to do it

inffy:
> > Actually might be better to just dismiss this one and go with the #45 which goes much deeper than this one :)
> 
> I thought about splitting my commit into several parts for convenience and adding this commit and #41, but haven't figured out how to do it

Smaller commits (one by one) would be much preferred vs a large merge 😄 

Not that familiar with git and don't really know how  easy it can be but I have  found this to be a great source of info:
https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

There is actually a section "Splitting a Commit" which might help

nisel11:
> > > Actually might be better to just dismiss this one and go with the #45 which goes much deeper than this one :)
> > 
> > 
> > I thought about splitting my commit into several parts for convenience and adding this commit and #41, but haven't figured out how to do it
> 
> Smaller commits would be much preferred vs a large merge 😄
> 
> Not that familiar with git and don't really know how easy it can be but I have found this to be a great source of info: https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History
> 
> There is actually a section "Splitting a Commit" which might help

I know how to do it, but I don't know what commits to split into, I'll try to look up how it is done in bluefin

PR #53, opened 2024-12-25
fix: remove pkcs11-provider workaround

ledif:
This pin is no longer needed and was removed in https://github.com/ublue-os/bluefin/pull/1993. Pulling in the same changes.


RealVishy (approved):
lgtm

NiHaiden (approved):

PR #81, opened 2025-01-04
fix(actions): try to recover if just fails to install

ledif:
Related to https://github.com/ublue-os/bluefin/pull/2106.

Lately, there are some flaky failures for jobs while installing `just`. For example, in the past few days, the following jobs to build images have failed:

- https://github.com/ublue-os/aurora/actions/runs/12607607121/job/35139162573
- https://github.com/ublue-os/aurora/actions/runs/12577208839/job/35054350351

The command `curl -L https://api.github.com/repos/casey/just/releases/latest | jq -r '.tag_name'` is sometimes returning "null" instead of an actual version, but it's difficult to tell if the curl call is failing or if it's returning a payload that is being parsed incorrectly.

In either case, adding `--retry` to curl and if it's still null, hardcoding a known good version should do the trick for now.

### DIFF
--- a/.github/workflows/generate-release.yml
+++ b/.github/workflows/generate-release.yml
@@ -36,14 +36,8 @@ jobs:
 
       - name: Install Just
         shell: bash
-        run: |
-          set -eoux pipefail
-          JUST_VERSION=$(curl -L https://api.github.com/repos/casey/just/releases/latest | jq -r '.tag_name')
-          curl -sSLO https://github.com/casey/just/releases/download/${JUST_VERSION}/just-${JUST_VERSION}-x86_64-unknown-linux-musl.tar.gz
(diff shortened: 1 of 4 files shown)


RealVishy (dismissed):

m2Giles:
Something like this would be a bit more resistant

```
          while [[ "${JUST_VERSION:-}" =~ null || -z "${JUST_VERSION:-}" ]]
          do
            JUST_VERSION=$(curl -L https://api.github.com/repos/casey/just/releases/latest | jq -r '.tag_name')
          done
```

ledif:
I updated the PR based on suggestions. Could I get a re-approval please? :)

m2Giles (approved):

PR #85, opened 2025-01-05
fix: artifacthub id

castrojo:
This uses the correct artifacthub id to list aurora on artifacthub. tulip is going to PR the metadata changes. 

Feel free to add yourselves as owners at the bottom but leave mine in there as I'm running the org account in artifacthub.

For those not familiar with artifacthub it's a place that indexes cloud native artifacts, here's an example of what it looks like: https://artifacthub.io/packages/container/achillobator/achillobator


ledif (approved):
Thanks!

PR #44, opened 2024-12-18
feat: openrazer-daemon

dreamyukii:
Add openrazer-daemon for configuring razer devices

PR #102, opened 2025-01-10
fix: make aurora-groups.sh executable 

dreamyukii:
forgot to make it executable so the service is failing

### DIFF
--- a/system_files/shared/usr/lib/systemd/system/aurora-groups.service
+++ b/system_files/shared/usr/lib/systemd/system/aurora-groups.service
@@ -1,10 +1,12 @@
 [Unit]
 Description=Add plugdev groups
+
 [Service]
 Type=oneshot
 ExecStart=/usr/libexec/aurora-groups
 Restart=on-failure
 RestartSec=30
 StartLimitInterval=0
+
 [Install]
 WantedBy=default.target
--- a/system_files/shared/usr/lib/systemd/system/aurora-groups.service
+++ b/system_files/shared/usr/lib/systemd/system/aurora-groups.service
@@ -3,7 +3,7 @@ Description=Add plugdev groups
 
 [Service]
 Type=oneshot
-ExecStart=/usr/libexec/aurora-groups
+ExecStart=/usr/libexec/aurora-groups.sh
 Restart=on-failure
 RestartSec=30
 StartLimitInterval=0
--- a/system_files/shared/usr/libexec/aurora-groups.sh
+++ b/system_files/shared/usr/libexec/aurora-groups.sh
@@ -24,7 +24,7 @@ append_group() {
 append_group plugdev
 
 wheelarray=($(getent group wheel | cut -d ":" -f 4 | tr  ',' '\n'))
-for user in $wheelarray
+for user in "${wheelarray[@]}"
 do
   usermod -aG plugdev $user
 done
--- a/system_files/shared/usr/libexec/aurora-groups.sh
+++ b/system_files/shared/usr/libexec/aurora-groups.sh
@@ -1,13 +1,16 @@
 #!/usr/bin/env bash
+
 # SCRIPT VERSION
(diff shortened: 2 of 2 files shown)


system_files/shared/usr/lib/systemd/system/aurora-groups.service
```diff
@@ -1,10 +1,12 @@
 [Unit]
 Description=Add plugdev groups
+
 [Service]
 Type=oneshot
 ExecStart=/usr/libexec/aurora-groups
```
ledif:
```suggestion
ExecStart=/usr/libexec/aurora-groups.sh
```
I think this should be the full path of the script. See https://github.com/ublue-os/aurora/issues/103 and https://github.com/ublue-os/aurora/pull/44#discussion_r1911033707

dreamyukii: according to [this](https://github.com/ublue-os/aurora/blob/main/system_files/dx/usr/lib/systemd/system/aurora-dx-groups.service) you don't need full path, and also i just check it the file is not executable so it can't be execute by systemd service

ledif: The service unit you linked works because the actual filename for the script it's calling is `/usr/libexec/aurora-dx-groups`. The script it is calling does not have an extension in its filename.

I created a custom build with your changes and the error is still there. The script is executable but your service unit isn't pointing to an actual file.

```
❯ systemctl status aurora-groups
● aurora-groups.service - Add plugdev groups
     Loaded: loaded (/usr/lib/systemd/system/aurora-groups.service; enabled; preset: disabled)
    Drop-In: /usr/lib/systemd/system/service.d
             └─10-timeout-abort.conf, 50-keep-warm.conf
     Active: activating (auto-restart) (Result: exit-code) since Fri 2025-01-10 18:29:01 CST; 2s ago
 Invocation: ca809057a56e41bfbe027aa28b5ba23b
    Process: 3434 ExecStart=/usr/libexec/aurora-groups (code=exited, status=203/EXEC)
   Main PID: 3434 (code=exited, status=203/EXEC)
   Mem peak: 1M
        CPU: 3ms

❯ cat /usr/lib/systemd/system/aurora-groups.service
[Unit]
Description=Add plugdev groups

[Service]
Type=oneshot
ExecStart=/usr/libexec/aurora-groups
Restart=on-failure
RestartSec=30
StartLimitInterval=0

[Install]
WantedBy=default.target

❯ ls -lah /usr/libexec/aurora-groups

ls: cannot access '/usr/libexec/aurora-groups': No such file or directory

~
❯ ls -lah /usr/libexec/aurora-groups.sh
-rwxr-xr-x. 2 root root 863 Dec 31  1969 /usr/libexec/aurora-groups.sh
```
When I changed the service unit to
```
ExecStart=/usr/libexec/aurora-groups.sh
```
The service works now. So you do indeed need the extension.

dreamyukii: i see, so it will not working without .sh

system_files/shared/usr/libexec/aurora-groups.sh
```diff
@@ -16,13 +19,16 @@ append_group() {
     grep "^$group_name:" /usr/lib/group | tee -a /etc/group > /dev/null
   fi
 }
+
 # Setup Groups
 append_group plugdev
+
 wheelarray=($(getent group wheel | cut -d ":" -f 4 | tr  ',' '\n'))
 for user in $wheelarray
```
jorti: To iterate the array you must use `${wheelarray[@]}`

ledif:
```suggestion
for user in "${wheelarray[@]}"
```

ledif (approved):
I think this should be good now.

PR #107, opened 2025-01-11
fix(just): fix configure-vfio recipe

ledif:
Port of my PR in Bluefin https://github.com/ublue-os/bluefin/pull/2140.


RealVishy (approved):
nice!

PR #123, opened 2025-01-14
chore: remove lxd from dx-groups ujust

inffy:
LXD was removed a while back but the ujust dx-group command still tries to add users to the lxd group and prints on error as no such group exists

This removes the lxd group from the ujust command


ledif (approved):
LGTM!

Issue #124, opened 2025-01-15
ublue-system-setup.service ran for >4m after an update due to unnecessary sudo calls


yacoob:
### Describe the bug

My laptop (Framework 13) took around 4.5min to execute `ublue-system-setup` during boot after an update. After some debugging I've found that this happened due to combination of two things:

1. `/usr/libexec/ublue-system-setup` which is run by this service uses `sudo` twice:
```shell
sudo tee /etc/modprobe.d/alsa.conf <<< "options snd-hda-intel index=1,0 model=auto,dell-headset-multi"
echo 0 | sudo tee /sys/module/snd_hda_intel/parameters/power_save
```

2. The service itself contains `Before=systemd-user-sessions.service`. 

This means each `sudo` call results in 2min timeout on `sudo` trying to talk to systemd and create a user session.

### What did you expect to happen?

This script doesn't need `sudo`, as it already executes as `root`. 

### Output of `bootc status`

```shell
This is on [yaurora](https://github.com/yacoob/yatomics/blob/main/recipes/yaurora.yaml) which is essentially aurora with a bunch of software added :) I'm pretty sure the problem is present in aurora too, although I haven't verified that first hand.


apiVersion: org.containers.bootc/v1
kind: BootcHost
metadata:
  name: host
spec:
  image:
    image: ghcr.io/yacoob/yaurora:latest
    transport: registry
    signature: containerPolicy
  bootOrder: default
status:
  staged: null
  booted:
    image:
      image:
        image: ghcr.io/yacoob/yaurora:latest
        transport: registry
        signature: containerPolicy
      version: '41.250114'
      timestamp: 2025-01-14T13:00:10Z
      imageDigest: sha256:288f2fec94f9d337e3455192319953c8f39694ead22fe12edfd4f251cdd3f936
    cachedUpdate: null
    incompatible: false
    pinned: false
    store: ostreeContainer
    ostree:
      checksum: ad0b8ff847b6988adb8f24523ef35c4a3a77ca5b80a404d03c6e39fb56909726
      deploySerial: 0
  rollback:
    image:
      image:
        image: ghcr.io/yacoob/yaurora:latest
        transport: registry
        signature: containerPolicy
      version: '41.241230'
      timestamp: 2024-12-30T13:56:23Z
      imageDigest: sha256:092d1777f37960de5688b2b62b5b40a612072b8324ee1d1d80cc088e565846ba
    cachedUpdate:
      image:
[...]

PR #125, opened 2025-01-15
fix: Update ublue-system-setup

inffy:
Fix for #124


inffy:
Need to check whats going on with the service file too

RealVishy (approved):

PR #135, opened 2025-01-17
feat: add org.gtk.Gtk3theme.Breeze for better GTK app integration

### LINKED ISSUE #132: Install org.gtk.Gtk3theme.Breeze flatpak by default
# Background 
Apps such as Firefox and Thunderbird don't seem to respect breeze theme without installing this package. This is one of those packages i install every time I install aurora. Unfortunately i do not have any screenshots but this is mostly notable when using Breeze color scheme with 'tint all windows' option that tints anything that uses Breeze theme with the selected accent colour. All i can say is that without this package, firefox and thunderbird do NOT look like this:

Image

inffy:
Fixes #132  

And removes adw-gtk flatpaks, we don't need those

### DIFF
--- a/aurora_flatpaks/flatpaks
+++ b/aurora_flatpaks/flatpaks
@@ -16,6 +16,4 @@ app/org.fedoraproject.MediaWriter/x86_64/stable
 app/io.missioncenter.MissionCenter/x86_64/stable
 app/org.gnome.DejaDup/x86_64/stable
 app/org.gnome.World.PikaBackup/x86_64/stable
-runtime/org.gtk.Gtk3theme.adw-gtk3/x86_64/3.22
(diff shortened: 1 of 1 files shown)


RealVishy (approved):

NiHaiden (approved):

etvt:
I had to apply these changes manually, I assume the installation of flatpaks only happens at first boot after install?  

It might be necessary to improve this mechanism for the users who are not regularly checking the repo changes, they should also somehow receive important system-wide changes / flatpaks. (I guess it might not be straightforward, for example removing anything could be problematic, maybe the user needs it for some reason and I don't know if it's possible to check if something was installed later by the user or by the system at first run.)

PR #141, opened 2025-01-21
feat(ci): Enable Docker Building, GHA Log Grouping

inffy:
Port from https://github.com/ublue-os/bluefin/pull/2114

Tested on VM but if you could also test this

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -2,8 +2,8 @@ repo_organization := "ublue-os"
 rechunker_image := "ghcr.io/hhd-dev/rechunk:v1.0.1"
 iso_builder_image := "ghcr.io/jasonn3/build-container-installer:v1.2.3"
 images := '(
-    [bluefin]=bluefin
-    [bluefin-dx]=bluefin-dx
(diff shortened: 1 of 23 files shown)


ledif (approved):
LGTM!

Justfile
```diff
@@ -148,8 +130,6 @@ build $image="aurora" $tag="latest" $flavor="main" rechunk="0" ghcr="0" pipeline
     # AKMODS Flavor and Kernel Version
     if [[ "${flavor}" =~ hwe ]]; then
         akmods_flavor="bazzite"
-    elif [[ "${tag}" =~ stable ]]; then
-        akmods_flavor="coreos-stable"
```
ledif, line 152: I believe these lines were not meant to be removed. I tracked down the discrepancy in the NVIDIA kernel module version vs the NVIDIA software stack versions (#148) to this removal. We are using the main kernel for akmods for `stable-daily` when we should be using the stable (coreos) kernel.

Issue #8, opened 2024-11-23
No graphical session after update to 41.20241117.3 on Aurora.


mejiasjrg:
### Describe the bug

The system updated from 41.20241112.1  to 41.20241117.3 and on the next boot it showed the arrow cursor for a little while and then a blank screen. It didn't even show the login screen. I changed to a terminal with _Ctrl-Alt-F2_ and could login with no issues. Tried _sudo systemctl restart sddm_, and got the blank screen again.
Rebooted the machine and selected the second entry in the grub menu, and the system booted normally to the graphical session. Then I issued _rpm-ostree rollback_, reboted, and issued _systemctl start rpm-ostreed-automatic.service_ to make the system update again, hoping there might be a new (different) update, but instead updated to 41.20241117.3 again, and again got the blank screen.
I finally did the rollback again and am posting from the system running 41.20241112.1

### What did you expect to happen?

I expected the system to boot normally running 41.20241117.3

### Output of `bootc status`

```shell
Current staged image: ghcr.io/ublue-os/aurora:stable
    Image version: 41.20241117.3 (2024-11-17 15:47:30 UTC)
    Image digest: sha256:fb3a023ce3c7609591c7524cb0c5a679453e5e0ff62cef6f334b4d7ca165965b
Current booted image: ghcr.io/ublue-os/aurora:stable
    Image version: 41.20241112.1 (2024-11-12 21:06:27 UTC)
    Image digest: sha256:f99e7cf789398600ffe7d1ade44ca0bceb31d76cbe8b8eb140a8b284804d6f6c
Current rollback image: ghcr.io/ublue-os/aurora:stable
    Image version: 41.20241117.3 (2024-11-17 15:47:30 UTC)
    Image digest: sha256:fb3a023ce3c7609591c7524cb0c5a679453e5e0ff62cef6f334b4d7ca165965b
```

### Output of `groups`

```shell
jm wheel
```

### Extra information or context

For the average user Aurora is intended to, this would have been a show-stopper. Not good.

mejiasjrg:
I've been fiddling with `aurora:stable`, but as I've had not much time to do it, this spans from Dec-12 to this day. Anyway, here is my report.

The system currently running:
❯ rpm-ostree status
State: idle
AutomaticUpdates: stage; rpm-ostreed-automatic.timer: no runs since boot
Deployments:
● ostree-image-signed:docker://ghcr.io/ublue-os/aurora:stable
[...]

PR #153, opened 2025-01-26
fix: use libheif from Fedora to support older Intel platforms

ledif:
Some context from Discord:
> @ledif
> - If we use negativo's libheif (current situation), we get better support for HEIF images (what is missing due to patents?) but old Intel hardware crashes
> - If we use Fedora's libheif, we lose some HEIF support (again, not sure what) but older Intel is supported
> ... Maybe we can just override the override directly in Aurora instead of starting a debate in the packages repo about what should be included by default for Bazzite as well

> @castrojo 
> yeah, do it in aurora
> I can almost guarantee bazzite won't turn off heif for people on i915

Related to https://github.com/ublue-os/aurora/issues/8.

### DIFF
--- a/build_files/base/05-override-install.sh
+++ b/build_files/base/05-override-install.sh
@@ -38,7 +38,7 @@ rm /etc/yum.repos.d/_copr_sentry-switcheroo-control_discrete.repo
 # https://github.com/ublue-os/aurora/issues/8
 rpm-ostree override replace \
     --experimental \
-    --from repo=updates \
+    --from repo=fedora \
         libheif heif-pixbuf-loader
 
 # Starship Shell Prompt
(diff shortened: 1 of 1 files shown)


inffy (approved):

NiHaiden (approved):

renner0e:
`:latest` image works on my affected hardware <3
@mejiasjrg

NiHaiden:
Whoop whoop!

> `:latest` image works on my affected hardware <3 @mejiasjrg

mejiasjrg:
> `:latest` image works on my affected hardware <3 @mejiasjrg

Likewise, @renner0e , my machine is running well on `41.20250209.1`. 
I had been running `aurora:stable` with the overrides until it updated last Sunday Feb 9th.
Thank you all for your work!

PR #160, opened 2025-01-29
chore: Replace solaar with flatpak

inffy:
Removes the solaar rpm from the base image and adds solaar flatpak


ledif (approved):
LGTM!

GarciaLnk:
I'd vouch for leaving out even the flatpak as users with logitech devices can install it themselves, and I imagine most of the install base don't need solaar.

PR #162, opened 2025-01-29
fix: rename tmpfiles.d conf file

ledif:
There's a space at the end of this filename. Not sure if that's causing issues but we should rename it in any case.


inffy (approved):
Tested on VM, works

NiHaiden (approved):

Issue #163, opened 2025-01-29
[Feature Request] Yubikey support


ledif:
Would be good to pull in https://github.com/ublue-os/bluefin/pull/2163.

PR #165, opened 2025-01-29
fix: allow wheel group users to access yubikeys

NiHaiden:
Porting over the change from Bluefin: 

https://github.com/ublue-os/bluefin/pull/2163

Issue: https://github.com/ublue-os/aurora/issues/163


NiHaiden:
Inffy was too fast.

PR #172, opened 2025-02-02
fix: flatpak comparison in ujust script

GarciaLnk:
The previous comparison approach didn't work as newlines in the list were trimmed to form one string with all the flatpaks, and comm compares line by line.

This uses bash arrays to properly manipulate the lists, additionally it sends a notification when new flatpaks are installed.

### DIFF
--- a/build_files/base/18-workarounds.sh
+++ b/build_files/base/18-workarounds.sh
@@ -24,10 +24,9 @@ mkdir -p /usr/share/ublue-os/aurora-cli
 cp /usr/share/ublue-os/bling/* /usr/share/ublue-os/aurora-cli
 
 # Copy flatpak list on image to compare against it on boot wo/ requiring curl to gh
-TOP_LEVEL_DIR=$(git rev-parse --show-toplevel)
-FLATPAK_LIST=($(cat ${TOP_LEVEL_DIR}/aurora_flatpaks/flatpaks))
+FLATPAK_LIST=($(curl https://raw.githubusercontent.com/ublue-os/aurora/main/aurora_flatpaks/flatpaks))
 if [[ ${IMAGE_NAME} =~ "dx" ]]; then
-    FLATPAK_LIST+=($(cat ${TOP_LEVEL_DIR}/dx_flatpaks/flatpaks))
+    FLATPAK_LIST+=($(curl https://raw.githubusercontent.com/ublue-os/aurora/main/dx_flatpaks/flatpaks))
 fi
 printf "%s\n" "${FLATPAK_LIST[@]}" > /usr/share/ublue-os/flatpak_list
 
--- a/build_files/base/18-workarounds.sh
+++ b/build_files/base/18-workarounds.sh
@@ -24,9 +24,9 @@ mkdir -p /usr/share/ublue-os/aurora-cli
 cp /usr/share/ublue-os/bling/* /usr/share/ublue-os/aurora-cli
 
 # Copy flatpak list on image to compare against it on boot wo/ requiring curl to gh
-FLATPAK_LIST=($(curl https://raw.githubusercontent.com/ublue-os/aurora/main/aurora_flatpaks/flatpaks))
+FLATPAK_LIST=($(cat /ctx/aurora_flatpaks/flatpaks))
 if [[ ${IMAGE_NAME} =~ "dx" ]]; then
-    FLATPAK_LIST+=($(curl https://raw.githubusercontent.com/ublue-os/aurora/main/dx_flatpaks/flatpaks))
+    FLATPAK_LIST+=($(cat /ctx/dx_flatpaks/flatpaks))
 fi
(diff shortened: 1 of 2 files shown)


just/aurora-system.just
```diff
@@ -216,27 +216,31 @@ configure-vfio ACTION="":
 [group('System')]
 [private]
 install-system-flatpaks CURR_LIST_FILE="":
-    #!/usr/bin/bash
+    #!/usr/bin/env bash
     CURR_LIST_FILE={{ CURR_LIST_FILE }}
     IMAGE_NAME=$(jq -r '."image-name"' < "/usr/share/ublue-os/image-info.json")
-    FLATPAK_LIST="$(curl https://raw.githubusercontent.com/ublue-os/aurora/main/aurora_flatpaks/flatpaks | tr '\n' ' ')"
-    flatpak --system -y install --or-update ${FLATPAK_LIST}
```
GarciaLnk, line 223: no clue how this got through lol

just/aurora-system.just
```diff
@@ -216,27 +216,31 @@ configure-vfio ACTION="":
...
+    FLATPAK_LIST=($(curl https://raw.githubusercontent.com/ublue-os/aurora/main/aurora_flatpaks/flatpaks))
```
etvt: Would it be possible / would it make sense to put these files in the image instead of downloading them every time from the internet, in order to:
1. avoid depending on internet connection to githubusercontent.com _on every boot_
2. only consider the historical version of the recommended flatpak list that was active at the point in time when the image was built.

GarciaLnk: Hmm, yeah we could curl the list in build time into /usr/share/ublue-os and compare against that in the recipe. I do like that approach much more, on it!

build_files/base/18-workarounds.sh
```diff
@@ -23,4 +23,11 @@ fi
 mkdir -p /usr/share/ublue-os/aurora-cli
 cp /usr/share/ublue-os/bling/* /usr/share/ublue-os/aurora-cli
 
+# Copy flatpak list on image to compare against it on boot wo/ requiring curl to gh
+FLATPAK_LIST=($(curl https://raw.githubusercontent.com/ublue-os/aurora/main/aurora_flatpaks/flatpaks))
+if [[ ${IMAGE_NAME} =~ "dx" ]]; then
+    FLATPAK_LIST+=($(curl https://raw.githubusercontent.com/ublue-os/aurora/main/dx_flatpaks/flatpaks))
+fi
+printf "%s\n" "${FLATPAK_LIST[@]}" > /usr/share/ublue-os/flatpak_list
+
```
ledif, line 32: Instead of grabbing these lists via curl, it makes sense to instead copy them in `build_base.sh` around: https://github.com/ublue-os/aurora/blob/a7f998d1a114c0b9c6c51a441389893059b34e7b/build_files/shared/build-base.sh#L56-L58

They are part of the repository and available as `/ctx/aurora_flatpaks/flatpaks` and `/ctx/dx_flatpaks/flatpaks`.

This also provides the benefit of testing/developing a branch with separate flatpaks from the `main` branch and being able to test that branch without pulling in flatpaks from the `main` branch.

GarciaLnk: Fixed, I forgot everything got mounted to /ctx

ledif (approved):
Thanks for the fixes. LGTM!

ledif:
I'm going to force merge this PR because it fixes an issue with code that has already been merged, so it needs to be available in the `stable` image. The current failing NVIDIA actions are unrelated.

just/aurora-system.just
```diff
@@ -216,27 +216,32 @@ configure-vfio ACTION="":
...
+            flatpak --system -y install --or-update "${FLATPAK_LIST[@]}"
         fi
     else
-        flatpak --system -y install --or-update ${FLATPAK_LIST}
+        IMAGE_NAME=$(jq -r '."image-name"' < "/usr/share/ublue-os/image-info.json")
+        FLATPAK_LIST=($(curl https://raw.githubusercontent.com/ublue-os/aurora/main/aurora_flatpaks/flatpaks))
+
+        if [[ ${IMAGE_NAME} =~ "dx" ]]; then
+            FLATPAK_LIST+=($(curl https://raw.githubusercontent.com/ublue-os/aurora/main/dx_flatpaks/flatpaks))
+        fi
```
etvt, line 243: Is there any reason why this else branch is not
`FLATPAK_LIST=($(cat /usr/share/ublue-os/flatpak_list))`
?

GarciaLnk: It keeps the old behavior of the recipe consistent. When the argument that points to a flatpak list is not given, it installs all the flatpaks from the repo (this only happens when the user manually does `ujust install-system-flatpaks`)

etvt: makes sense, thanks!

PR #173, opened 2025-02-03
fix: override mesa packages and add nvidia repo to fix nvidia builds

GarciaLnk:
this should fix the current build issues on nvidia images


ledif:
Adding context from Discord:

> @m2Giles 
> Looking to fix the actual script
> https://github.com/ublue-os/hwe/pull/314

Source: https://discord.com/channels/1072614816579063828/1072617059265032342/1335781548523524149

inffy:
This was handled with #174

PR #174, opened 2025-02-03
chore: un-cliwrap and use dnf5 instead of rpm-ostree

inffy:
I feel pain...

ported from https://github.com/ublue-os/bluefin/pull/2143/commits

If you could test and look this over. I will test on a VM now.


build_files/base/01-build-fix.sh
m2Giles, line 1: This whole file can be removed.

build_files/base/03-install-kernel-akmods.sh
```diff
@@ -58,7 +65,7 @@ if [[ "${IMAGE_NAME}" =~ nvidia ]]; then
     mv /tmp/rpms/* /tmp/akmods-rpms/
 
     # Install Nvidia RPMs
-    curl -Lo /tmp/nvidia-install.sh https://raw.githubusercontent.com/ublue-os/hwe/main/nvidia-install.sh
+    curl -Lo /tmp/nvidia-install.sh https://raw.githubusercontent.com/ublue-os/hwe/b3a3dbddf4af81cfbfa7526c1918c9b9f014f86b/nvidia-install.sh # Change when nvidia-install.sh updates
```
m2Giles: This has been merged so you don't need to edit this.

build_files/base/07-base-image-changes.sh
```diff
@@ -4,11 +4,20 @@ echo "::group:: ===$(basename "$0")==="
 
 set -ouex pipefail
 
+
 # Restore x11 for Nvidia Images
 if [[ "${FEDORA_MAJOR_VERSION}" -eq "40" ]]; then
     rpm-ostree install plasma-workspace-x11
 fi
 
+# Remove desktop entries
```
m2Giles, line 7: Pretty sure you guys don't have these files.

build_files/base/17-cleanup.sh
m2Giles, line 1: We will probably want to use dnf5 config-manager setopt to turn these off instead of using sed.

build_files/shared/build-base.sh
```diff
@@ -2,7 +2,26 @@
 
 set -eou pipefail
 
+
 echo "::group:: Copy Files"
+
+# there is no 'rpm-ostree cliwrap uninstall-from-root', but this is close enough. See:
+# https://github.com/coreos/rpm-ostree/blob/6d2548ddb2bfa8f4e9bafe5c6e717cf9531d8001/rust/src/cliwrap.rs#L25-L32
+if [ -d /usr/libexec/rpm-ostree/wrapped ]; then
```
m2Giles, line 10: Good this has a guard.

build_files/base/05-override-install.sh
```diff
@@ -4,10 +4,10 @@ echo "::group:: ===$(basename "$0")==="
 
 set -eoux pipefail
 
+
 # Patched shell
-rpm-ostree override replace \
-    --experimental \
-    --from repo=copr:copr.fedorainfracloud.org:ublue-os:staging \
+dnf5 -y swap \
```
m2Giles, line 9: dnf5 sway only takes 2 specs. The orginal and the one you want to replace it with.

For here this should change to

```bash
dnf5 -y swap \
    kf6-kio-core kf6-kio-core
```

GarciaLnk:
We should not merge this until uupd is ready.

NiHaiden (approved):

PR #175, opened 2025-02-04
feat: remove inputleap from base image

inffy:
Removes inputleap package from the image and replaces it with the flatpak version.

Also removes filelight flatpak, as its already part of the image and it works better than the flatpak which need extra perms to work

### DIFF
--- a/system_files/shared/usr/libexec/ublue-flatpak-manager
+++ b/system_files/shared/usr/libexec/ublue-flatpak-manager
@@ -1,7 +1,7 @@
 #!/usr/bin/bash
 
 # Script Version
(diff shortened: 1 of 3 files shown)


system_files/shared/usr/libexec/ublue-flatpak-manager
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/bash
 
 # Script Version
-VER=2
+VER=3
```
GarciaLnk: we don't have to increment the VER number when we change the flatpak list, that runs on its own path before the VER check

PR #181, opened 2025-02-05
fix: make ujust update use direct calls to rpm-ostree and others

inffy:


NiHaiden (approved):

system_files/shared/usr/share/ublue-os/just/10-update.just
```diff
@@ -5,9 +5,9 @@ alias upgrade := update
 # Update system, flatpaks, and containers all at once
 update:
     #!/usr/bin/bash
-    TOPGRADE_CONFIG="/usr/share/ublue-os/topgrade"
-    /usr/bin/grep "^LockLayering=true" /etc/rpm-ostreed.conf &>/dev/null && TOPGRADE_CONFIG="${TOPGRADE_CONFIG}-bootc"
-    /usr/bin/topgrade --config "${TOPGRADE_CONFIG}.toml" --keep
+    rpm-ostree upgrade
+    flatpak update -y
+    brew update
```
etvt, line 10: If I'm not mistaken, the toml file contains also `distrobox` and `firmware`, shouldn't they be added here?  
Also, I thought brew is updated through a timer and was removed from `ujust update` at some point. (But maybe could be useful to trigger it here explicitly.)

inffy: Distroboxes and fwupd were dropped. Atleast for now

PR #187, opened 2025-02-05
fix: make cursor smaller in gtk flatpaks when fractional scaling is enabled

GarciaLnk:
This works with the breeze cursor, ymmv with other cursor themes

Before:
before

After:
after


inffy (changes requested):
Would be maybe better to put in 
system_files/shared/usr/libexec/ublue-user-setup where all other overrides are. And then upp the version number one up

GarciaLnk:
done, I've also changed the overrides to be user-wide since that script is run through a user service

inffy (approved):

PR #189, opened 2025-02-06
fix: Justfile to avoid collision with fedora flatpaks

NiHaiden:
This very small added flag ensures we don't collide with fedora flatpaks that might be installed by a user and just reinstalls them from scratch from flathub.

Credits go to @Zeglius <3.


inffy (approved):
LGTM

PR #200, opened 2025-02-08
feat: cleanup fonts,sddm folder and add gpu-screen-recorder to gaming flatpak just recipe

dreamyukii:
cleanup fonts and stuff


packages.json
```diff
@@ -45,6 +45,7 @@
 				"ptyxis",
 				"rclone",
 				"restic",
+				"rsms-inter-fonts",
```
inffy, line 48: I was thinking do we actually need even the Inter fonts. I guess Bluefin uses these for the Gnome interface and apps but do we actually need these in KDE.

dreamyukii: also ubuntu fonts, i think some of fonts can be remove safely if nobody use that in Aurora

inffy: Of course hard to say if someone does use them but by default I think nothing uses those.

inffy:
@NiHaiden @ledif @RealVishy What do you guys think? I think we could really slim down on the fonts we have on the image.

We have embellish and font downloader for people if they need specific fonts..

NiHaiden:
> @NiHaiden @ledif @RealVishy What do you guys think? I think we could really slim down on the fonts we have on the image.
> 
> We have embellish and font downloader for people if they need specific fonts..

I think we can safely remove the Fonts from the system and include something else that will download the fonts people want, like Font Downloader from Flathub. (https://flathub.org/apps/org.gustavoperedo.FontDownloader)

NiHaiden:
But Inter Fonts can be kept i think, they might be one of the more used fonts for people.

inffy:
> > @NiHaiden @ledif @RealVishy What do you guys think? I think we could really slim down on the fonts we have on the image.
> > We have embellish and font downloader for people if they need specific fonts..
> 
> I think we can safely remove the Fonts from the system and include something else that will download the fonts people want, like Font Downloader from Flathub. (https://flathub.org/apps/org.gustavoperedo.FontDownloader)

Yes, Font Downloader flatpak is already on the image, and -dx image has also Embellish for managing Nerd fonts.

packages.json
ledif: Just a data point from me. I am using the Inter fonts as my default font which I installed in my home directory. I think it would be nice to include this by default :smile:.

ledif:
This is the current SDDM theme that already has the latest logo:

image

If we look at `/usr/share/sddm/themes/01-breeze-fedora/theme.conf`, it shows that the logo is actually coming from `/usr/share/pixmaps/fedora_whitelogo.svg`. This is already set to the new logo.

```
$ cat /usr/share/sddm/themes/01-breeze-fedora/theme.conf
[General]
showlogo=shown
showClock=true
logo=/usr/share/pixmaps/fedora_whitelogo.svg
type=image
color=#1d99f3
fontSize=10
background=/usr/share/backgrounds/default.png
needsFullUserModel=false
```

So I think the file [system_files/kinoite/usr/share/sddm/themes/01-breeze-fedora/default-logo.svg](https://github.com/ublue-os/aurora/pull/200/files#diff-fab5f48f76815b539edacf60cee4ddd1d99a690964433faceeb5d0467bc45c68) can simply just be removed.

dreamyukii:
> This is the current SDDM theme that already has the latest logo:
> 
> image
> 
> If we look at `/usr/share/sddm/themes/01-breeze-fedora/theme.conf`, it shows that the logo is actually coming from `/usr/share/pixmaps/fedora_whitelogo.svg`. This is already set to the new logo.
> 
> ```
> $ cat /usr/share/sddm/themes/01-breeze-fedora/theme.conf
> [General]
> showlogo=shown
> showClock=true
> logo=/usr/share/pixmaps/fedora_whitelogo.svg
> type=image
> color=#1d99f3
> fontSize=10
> background=/usr/share/backgrounds/default.png
> needsFullUserModel=false
> ```
> 
> So I think the file [system_files/kinoite/usr/share/sddm/themes/01-breeze-fedora/default-logo.svg](https://github.com/ublue-os/aurora/pull/200/files#diff-fab5f48f76815b539edacf60cee4ddd1d99a690964433faceeb5d0467bc45c68) can simply just be removed.

done

RealVishy (dismissed):
I'm happy to remove the fonts now that we have fontdownloader/embellish. Let's ask on the discourse and then PR removing the fonts it if it's popular?

NiHaiden (approved):

PR #204, opened 2025-02-09
fix: add 23.08 runtimes for mangohud and gamescope for lutris/heroic

### LINKED ISSUE #203: bug: wrong version of gamescope and mangohud bundled with gaming ujust
Lutris and Heroic require the 23.08 branches of gamescope and mangohud for the options to work in their GUIs.

We should add those versions to the ujust and add a comment so we can keep track of the versions they need.

inffy:
Fixes #203


RealVishy (approved):
lgtm

Issue #208, opened 2025-02-10
Pre-installed flatpaks return after reboot after uninstalling them


n3thshan:
### Describe the bug

when i uninstalled the pre-installed apps like Mission Center, pika backup, ktailctl, etc.... via KDE discover or Warehouse, they uninstall at first instance. But upon rebooting, they return. This behaviour is not present with user installed apps, that is, apps "I" install from discover like anki, obsidian, etc... For example, i just installed upscaler, uninstalled it and also uninstalled font downloader (pre-installed). Upon rebooting, upscaler was missing (as expected) but font downloader was there. 

### What did you expect to happen?

Pre-installed flatpaks should be removed upon uninstalling them regardless of a reboot or update.

### Output of `bootc status`

```shell
Booted ostree
           Commit: 7e86839b5392cc1b7aabbafe61f2601bff247f939335d79c51db6d24457ef349

  Rollback image: ghcr.io/ublue-os/aurora:stable
          Digest: sha256:48560c2aa9fbc6a2e1a1c21fc25ec7bd164e17020826d2a41cf36f00353a4424
         Version: 41.20250209.1 (2025-02-09 05:57:01 UTC)
```

### Output of `groups`

```shell
nethshan wheel plugdev
```

inffy:
Maybe this is because #170 and/or #172 

@GarciaLnk any ideas?

GarciaLnk:
Ugh, I was passing the ujust arguments incorrectly, it should be `ujust install-system-flatpaks "${LOCAL_UBLUE_DIR}/flatpak_current_list"` not `ujust install-system-flatpaks CURR_LIST_FILE="${LOCAL_UBLUE_DIR}/flatpak_current_list"`...

https://github.com/ublue-os/aurora/issues/210 is also related to this I'd imagine

Issue #210, opened 2025-02-10
ublue-flatpak-manager fails to run


etvt:
### Describe the bug

Realized none of the new flatpaks got installed that are listed in `aurora_flatpaks/flatpaks` or `dx_flatpaks/flatpaks`, so upon checking the logs of the service found this:

```
Feb 10 09:02:54 hostname systemd[2181]: Starting ublue-flatpak-manager.service - Manage flatpaks...
Feb 10 09:02:54 hostname ublue-flatpak-manager[2250]: /run/user/1000/just/just-0BYEfW/install-system-flatpaks: line 505: CURR_LIST_FILE=/home/username/.local/share/ublue/flatpak_current_list: No such file or directory
Feb 10 09:02:54 hostname ublue-flatpak-manager[2252]: Looking for matches…
Feb 10 09:02:54 hostname ublue-flatpak-manager[2252]: F: An error was encountered searching remote ‘flathub’ for ‘app/org.mozilla.Thunderbird/x86_64/stable’: Unable to load summary from remote flathub: While fetching https://dl.flathub.org/repo/summary.idx: [6] Could not resolve hostname
Feb 10 09:02:54 hostname ublue-flatpak-manager[2252]: error: No remote refs found for ‘app/org.mozilla.Thunderbird/x86_64/stable’
Feb 10 09:02:54 hostname ublue-flatpak-manager[2230]: error: Recipe `install-system-flatpaks` failed with exit code 1
Feb 10 09:02:54 hostname ublue-flatpak-manager[2220]: Flatpak manager v2 has already ran. Exiting...
Feb 10 09:02:54 hostname systemd[2181]: Finished ublue-flatpak-manager.service - Manage flatpaks.
```

Not sure if `No such file or directory` is a real error or just a warning.
Regarding `Could not resolve hostname`, maybe the script runs too early, when wlan is not yet connected?

I got the same errors consistently in all the previous days.

### What did you expect to happen?

Flatpak manager should run successfully and install the new missing flatpaks.

### Output of `bootc status`

```shell
● Booted image: ghcr.io/ublue-os/aurora-dx:stable-daily
        Digest: sha256:edbe2ee266688fcdbd337b8c3673b070e7205bd2e44e1abeefdda9ad8c736a6a
       Version: 41.20250210.4 (2025-02-10 07:25:28 UTC)

  Rollback image: ghcr.io/ublue-os/aurora-dx:stable-daily
[...]

PR #211, opened 2025-02-10
fix: correctly pass argument to `ujust install-system-flatpaks`

GarciaLnk:
Should fix https://github.com/ublue-os/aurora/issues/208 and https://github.com/ublue-os/aurora/issues/210


inffy (approved):

GarciaLnk:
Actually, not sure about https://github.com/ublue-os/aurora/issues/210 since it also reports an error resolving flathub dns, even though `ublue-flatpak-manager` runs after `network-online.target`

inffy:
> Actually, not sure about #210 since it also reports an error resolving flathub dns, even though `ublue-flatpak-manager` runs after `network-online.target`

That sounds more like a network/dns issue. Kind of strange but maybe the network just wakes slowly.

PR #213, opened 2025-02-13
feat: add starship config

dreamyukii:
starshiptheme

add starship config

### DIFF
--- a/system_files/kinoite/etc/skel/.config/starship.toml
+++ b/system_files/kinoite/etc/skel/.config/starship.toml
@@ -1,0 +1,25 @@
+# Get editor completions based on the config schema
+"$schema" = 'https://starship.rs/config-schema.json'
+
+add_newline = true
+
+[line_break]
+disabled = true
+
+[username]
+style_user = 'blue bold'
+style_root = 'red bold'
+format = ' [$user]($style)'
+disabled = false
+show_always = true
+
+[hostname]
+ssh_only = false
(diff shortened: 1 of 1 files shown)


inffy:
Im not very familiar with starships toml configs, but personally i would prefer the actualy "command line" be on a new row like

[username] at [hostname] [dir]
-> [command prompt]

I guess its this line?

[line_break]
disabled = true

I'll add @NiHaiden @ledif @RealVishy to have more comments

RealVishy:
> Im not very familiar with starships toml configs, but personally i would prefer the actualy "command line" be on a new row like

Same, i'm kinda used to that from powerlevel10k

Here's what it looks like with the line_break enabled.

image

inffy:
Tested it myself too with some changes:

image

if the line_break is set to true, then the empty space maybe should be removed from the format = ' [user]($style)´

inffy:
[starship.toml.txt](https://github.com/user-attachments/files/18798718/starship.toml.txt)


Here is a one config i played around with (added nerd fonts icons, small layout changes)

dreamyukii:
> [starship.toml.txt](https://github.com/user-attachments/files/18798718/starship.toml.txt)
> 
> Here is a one config i played around with (added nerd fonts icons, small layout changes)

this looks great

RealVishy:
we should PR a message to the motd so it tells people they can change the prompt by changing the starship.toml

inffy (approved):

PR #222, opened 2025-02-17
add dependencies required for game development with the Bevy game engine

FreddyFunk:
Docs can be found here:

https://github.com/bevyengine/bevy/blob/latest/docs/linux_dependencies.md#Fedora

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -82,6 +82,7 @@
 			"dx": [
 				"adobe-source-code-pro-fonts",
 				"android-tools",
+				"alsa-lib-devel",
 				"bcc",
 				"bpftop",
 				"bpftrace",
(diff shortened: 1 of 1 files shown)


etvt:
Just out of curiosity, cannot Bevy be used in a distrobox / devcontainer (with wayland/device forwarding)?

ledif:
In my opinion, it does not make sense to add development dependencies into the base image. The preferred method would be to use a distrobox for the project and add Bevy's dependencies there. Since they're only dev dependencies, they shouldn't be needed to actually execute the resulting program built with Bevy and you should be able to run it on the host.

inffy:
These don't belong to base images in my opinion.

PR #227, opened 2025-02-21
chore: small cleanup

dreamyukii:
small cleanup like using dnf config-manager addrepo for openrazer repository

### DIFF
--- a/build_files/base/05-override-install.sh
+++ b/build_files/base/05-override-install.sh
@@ -57,10 +57,6 @@ curl --retry 3 -Lo /usr/share/bash-prexec https://raw.githubusercontent.com/rcal
 # Consolidate Just Files
 find /tmp/just -iname '*.just' -exec printf "\n\n" \; -exec cat {} \; >> /usr/share/ublue-os/just/60-custom.just
 
-# Register Fonts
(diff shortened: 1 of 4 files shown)


build_files/base/05-override-install.sh
```diff
@@ -57,10 +57,6 @@ curl --retry 3 -Lo /usr/share/bash-prexec https://raw.githubusercontent.com/rcal
 # Consolidate Just Files
 find /tmp/just -iname '*.just' -exec printf "\n\n" \; -exec cat {} \; >> /usr/share/ublue-os/just/60-custom.just
 
-# Register Fonts
-fc-cache -f /usr/share/fonts/ubuntu
-fc-cache -f /usr/share/fonts/inter
```
ledif, line 62: We don't need the ubuntu line because we removed the Ubuntu fonts, but do we still need the inter line? I see that we still install the `rsms-inter-fonts` package.

dreamyukii: i think this is only needed if you install font manually by copy and paste into `/usr/share/fonts`, since we switch to rpm package this is not needed.

ledif (approved):
LGTM

PR #228, opened 2025-02-21
chore: remove alternatives workaround

inffy:
Ported from https://github.com/ublue-os/bluefin/pull/2215/

main has removed the workaround since upstream should have fixed it.

Relates: https://github.com/ublue-os/main/pull/721


ledif (approved):
LGTM

PR #234, opened 2025-02-23
chore: use `ublue-os/packages` for ublue-related packages

inffy:
From https://github.com/ublue-os/bluefin/pull/2218


ledif (approved):
LGTM. I see that it's not actually using the ublue-os:packages repo when installing packages, but I think that's intended and the idea is that it will start using it once they are removed from ublue-os:staging.

inffy:
> LGTM. I see that it's not actually using the ublue-os:packages repo when installing packages, but I think that's intended and the idea is that it will start using it once they are removed from ublue-os:staging.

That is how I understood it that the packages are going to be removed from staging repo in the coming days

PR #236, opened 2025-02-24
fix: update package list to only include symbol nerd fonts

NiHaiden:
This fixes mutliple reports of the iso size increasing by a lot because nerd fonts pull in about 1 gigabyte worth of data from the terra repo.


GarciaLnk:
we should remove `che/nerd-fonts` copr while we are at it

inffy (approved):

PR #245, opened 2025-02-27
chore: pin latest to kernel 6.12

inffy:
This pins aurora latest kernels to our newest good builds:

    main/nvidia flavors: 6.12.15
    hwe/hwe-nvidia flavors: 6.12.12.bazzite

These are the newest known good kernels for the respective flavors.

The purpose of the pin is avoid shipping kernel 6.13 until the major issues are resolved (eg "bad page state" and fuse issues crashing systems and flatpaks).


inffy:
Looks like this builds properly so this should be merged once check are done.

The bad 6.13.4 version is propably landing on ublue on tonights builds. HWE images are actually been churning on 6.13 for a while now, but those are low consumables so not many users.

6.13.5 should include the fix, but it has not been pushed to testing yet. So better to be safe.

NiHaiden (approved):

NiHaiden:
Thank you for doing this, @inffy!

PR #270, opened 2025-03-14
feat: add container detection to dnf

ledif:
Porting https://github.com/ublue-os/bazzite/pull/2357 and also replacing `yum` with a link to `dnf`.


ledif:
Tested in a VM and for some reason the symbolic link isn't getting created. I might just need to make changes in both `dnf` and `yum`.

Screenshot_20250314_105957

ledif:
I removed the symlink and just made the same changes to yum. Should be good to go now.

inffy (approved):

PR #277, opened 2025-03-16
feat: tell discover to only show flatpaks

NiHaiden:
This adds a one-line change to the build files that will alter the Discover Desktop Shortcut with a command-line argument so that it only shows Flatpaks instead of Flatpaks, Themes and other stuff.


inffy (approved):
LGTM

PR #286, opened 2025-03-20
chore: remove DevPod rpm

RealVishy:
Won't work until the flatpak is actually published on flathub


NiHaiden (approved):

inffy:
I think we can just remove the devpod rpm now. `stable` will propably stay on F41 for a week or two. F42 can manage until the flatpak is ready (as the rpm had lot of issues sometimes)

PR #309, opened 2025-04-02
chore: remove hidden system-flatpak-install from the ujust list

inffy:
I don't think this should be hidden nowadays anymore


ledif (approved):
LGTM

RealVishy (approved):
ya, makes sense for people rebasing

PR #313, opened 2025-04-05
fix: Revert "Flatpak installation script missing Flathub configuration "

inffy:


L0g0ff:
If you're coming from another build without Flathub integration (whether due to rebasing or using a custom ISO generated with 'sudo bluebuild generate-iso --iso-name aurora-dx-hwe.iso image ghcr.io/ublue-os/aurora-dx-hwe'), removing this fix will cause the install-system-flatpak process to fail.

PR #315, opened 2025-04-07
chore: simplify packages.sh install scripts

inffy:
This cleans up the packages.json removing the kinoite objects. Also needed to do package exclusions as some packages are still missing from F42.

Also adds section for F42.


ledif (approved):
LGTM. Hopefully those 41-only packages get updated to 42 soon.

PR #322, opened 2025-04-09
feat: Add samsung unified linux driver (#2321)

NiHaiden:
Picked from bluefin, we should include this too. 
https://github.com/ublue-os/bluefin/pull/2321


inffy (approved):
Love me some printers

RealVishy (approved):

PR #333, opened 2025-04-12
feat: add WIP live ISO workflow 

inffy:
Maybe we want this too?


p5:
Unfortunately readymade does not (currently?) have any way to create users, which may be a blocker for Aurora at the moment

inffy:
> Unfortunately readymade does not (currently?) have any way to create users, which may be a blocker for Aurora at the moment

ah yeah i just looked quickly but didn't see that. Yes this is an issue then for us then

PR #334, opened 2025-04-12
chore: add gates for f42 launch

inffy:
Seems the switcheroo-control issue is still there and F42 launches on tuesday, so we need some gating that F42 builds won't try to swap the switcheroo-control package to the terra one and break builds once `latest` is moved to F42


inffy:
I think this is fine for now

NiHaiden (approved):

PR #365, opened 2025-04-20
feat: add taidan as first boot OOBE experience

tulilirockz:
This allows us to actually use Readymade as our installer on Aurora :)


inffy:
Taidan still isn't on Terra

tulilirockz:
Taidan will be available once we do this: https://github.com/ublue-os/packages/pull/405

PR #366, opened 2025-04-21
feat: enable ganto copr for incus

inffy:
Enable ganto copr for incus etc for Fedora 42


NiHaiden (approved):

inffy:
Can't use this yet. There is no available builds for F42

PR #368, opened 2025-04-23
fix(hwe): switch to skip-unavailable

inffy:
surface is still missing some packages, switch to skipping them until they are available but enable builds

https://github.com/ublue-os/bluefin/pull/2378


RealVishy (approved):
TIL that was a flag in dnf5. Good to start shipping f42 to hwe though!

PR #374, opened 2025-04-24
fix(flatpak): remove fedora flatpak service and repos

inffy:
Seems that fedora is adding a service to enable fedora  flatpak repos

### DIFF
--- a/build_files/base/17-cleanup.sh
+++ b/build_files/base/17-cleanup.sh
@@ -39,6 +39,11 @@ for file in fish htop nvtop; do
     fi
 done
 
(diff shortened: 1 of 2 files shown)


build_files/base/17-cleanup.sh
```diff
@@ -39,6 +39,11 @@ for file in fish htop nvtop; do
     fi
 done
 
+#Add the Flathub Flatpak remote and remove the Fedora Flatpak remote
+flatpak remote-delete --system fedora
```
p5:
```suggestion
# Add the Flathub Flatpak remote
```
This remote isn't actually part of the image

NiHaiden (approved):

PR #375, opened 2025-04-24
fix: devpod flatpak fix for nvidia

inffy:
devpod flatpak has issue with nvidia, this sets env override for the flatpak devpod app

### DIFF
--- a/system_files/shared/usr/libexec/ublue-user-setup
+++ b/system_files/shared/usr/libexec/ublue-user-setup
@@ -1,7 +1,7 @@
 #!/usr/bin/bash
 
 # SCRIPT VERSION
-USER_SETUP_VER=10
+USER_SETUP_VER=11
 USER_SETUP_VER_FILE="${XDG_DATA_HOME:-$HOME/.local/share}/ublue/user-setup"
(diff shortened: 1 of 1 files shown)


NiHaiden (approved):

system_files/shared/usr/libexec/ublue-user-setup
```diff
@@ -69,6 +69,9 @@ flatpak override --user --unset-env=QT_QPA_PLATFORMTHEME
 # smaller cursor on GTK apps - should be fixed by F42 with GTK 4.17
 flatpak override --user --env=USE_POINTER_VIEWPORT=1
 
+# webkit override to devpod for nvidia users
+flatpak override --user --env=WEBKIT_DISABLE_COMPOSITING_MODE=1 sh.loft.devpod
```
p5, line 73: What will this do if not on a DX image, or if Devpod has been removed?

inffy: Tried with removing the flatpak and then adding that environment var, it doesn't give any errors so I think it should be fine?

inffy: after installing devpod back, the env var is added to the sh.loft.devpod flatpak

PR #377, opened 2025-04-24
fix: remove incus alias

goek-han:
It is unnecessary and inconsistent with the other install commands which have no alias.


tulilirockz (approved):
Looks good to me, makes stuff more consistent too

PR #380, opened 2025-04-25
feat: integrate firewall-config as a kcm

fesdonomist:
Adds an external KCM that opens `firewall-config` in System Settings.

image


inffy (approved):
Lgtm very nice

PR #381, opened 2025-04-25
fix: make samba usershares work OOTB

fesdonomist:
Sets up SELinux and firewalld so that creating a Samba usershare through Dolphin (right click a folder -> Properties -> Share) works without silently failing, as well as preventing Samba from automatically exporting users' home folder as a share.


RealVishy (approved):
lgtm

PR #383, opened 2025-04-25
fix(rebase-helper): Guard rebase helper

inffy:
Adds small guards to rebase-helper ujust.

Until stable gets Fedora 42, if the user is on `latest` already, not show other streams.

IF on `stable or stable-daily` then allow all but add a note that latest wont allow rollback until stable is f42


hanthor:
I'll probably do that too. I figured we wouldn't merge until we increment to F42

inffy:
> I'll probably do that too. I figured we wouldn't merge until we increment to F42

Yeah our situation is little different as we don't have GTS (and propably will not have one soon).

PR #384, opened 2025-04-25
chore(iso): drop legacy nvidia ISOs

inffy:
Drop building the legacy-nvidia driver ISOs.

We are currently the only ones that build these in  ublue. 

So keep building images for as long as nvidia allows, but drop the ISOs. Saves builders, bandhwith and bucket costs

Needs the OK from Niklas, as needs also changes to homepage image picker


NiHaiden (approved):
Thank you for dropping these, its time we move on from the legacy stuff somewhat. If people really need the legacy nvidia images, they can rebase anyway. 
I will update the website accordingly. 

LGTM.

PR #385, opened 2025-04-25
fix: cleanup dockerenv in brew.sh

renner0e:
This blows up local/non-rechunked on F42 builds as the system is recognized as a container otherwise
image


inffy (approved):

tulilirockz:
https://github.com/ublue-os/aurora/pull/387 should make this not necessary anymore

PR #387, opened 2025-04-25
chore: cleanup everything brew-related that ublue-brew entirely covers

tulilirockz:
Same PR from https://github.com/ublue-os/bluefin/pull/2420


inffy:
We need to remove this right?

https://github.com/ublue-os/aurora/blob/301e058a0b9cf0bc1f63f74e5c23fd5ddf7bfd00/build_files/base/17-cleanup.sh#L21

PR #388, opened 2025-04-26
feat: Add python-tkinter

TeamLinux01:
Allows python apps to use import tkinter to run GUI python apps.


inffy:
Is there some specific reason this needs to be on image and not in brew or a distrobox?

TeamLinux01:
It would make it easier to run apps that import it out-of-the-box.
Discussed this in discord and it sounds like it should just be added via container.

Thanks for the discussion.

PR #401, opened 2025-04-29
fix: /etc/hostname fixup

inffy:
Previous solution could overwrite users hostname. This creates an empty file if it  doesn't exist


ledif:
Doesn't this service not run if `/etc/hostname` doesn't exist though, according to this line: 

```
ConditionPathExists=!/etc/hostname
```

inffy:
> ConditionPathExists

atleast the systemd man pages say this

``` If the absolute path name passed to ConditionPathExists= is prefixed
with an exclamation mark ("!"), the test is negated, and the
unit is only started if the path does not exist.
```

PR #414, opened 2025-05-02
fix(uupd): disable distrobox module from uupd

inffy:
Disables updating distroboxes during uupd run.

Needs the newest uupd, so drafting until build is done


inffy:
uupd pkg bumped in copr, so this is good to go

PR #425, opened 2025-05-04
fix: remove kns + fwupd backends from Discover more reliably

fesdonomist:
This works no matter which way you launch Discover, including when you right click to uninstall something or run it from the terminal.
Doesn't seem to have any adverse side effects, it just doesn't register them as backends anymore.


fesdonomist:
https://src.fedoraproject.org/fork/renner/rpms/plasma-discover/c/31bc11d9c7d9ebdbd63cc67a12671b460aa2b689?branch=split-fwupd

this would be better if/when it gets merged

renner0e:
If this gets merged upstream then I'll do the kns stuff as well
https://src.fedoraproject.org/rpms/plasma-discover/pull-request/27

build_files/base/07-base-image-changes.sh
```diff
@@ -28,8 +28,8 @@ sed -i 's@Exec=ptyxis@Exec=kde-ptyxis@g' /usr/share/applications/org.gnome.Ptyxi
 sed -i 's@Keywords=@Keywords=konsole;console;@g' /usr/share/applications/org.gnome.Ptyxis.desktop
 cp /usr/share/applications/org.gnome.Ptyxis.desktop /usr/share/kglobalaccel/org.gnome.Ptyxis.desktop
 
-# Plasma Discover
-sed -i 's/^Exec=plasma-discover/& --backends flatpak-backend/' /usr/share/applications/org.kde.discover.desktop
```
tulilirockz, line 32: Do we want to remove the `--backends` thing?

ledif (approved):
LGTT (looks good to @tulilirockz)

tulilirockz (approved):
LGTT (looks good to @tulilirockz)

build_files/base/07-base-image-changes.sh
fesdonomist: Yes, none of the backends will register if they're completely deleted so this is redundant

PR #472, opened 2025-05-15
feat: add polkit rules pkg

inffy:


tulilirockz:
Should be a drop-in thing, I think this is fine as-is

PR #479, opened 2025-05-17
feat: replace gtk3 firewall with plasma-firewall

renner0e:
replaces this GTK firewall:
image

with that:
image


tulilirockz (approved):

tulilirockz:
This firewall tool is super dumb, we dont include it on bluefin, and it seems to be deprecated. Whatever gets rid of GTK stuff on Aurora, is best. Hell yeah!

GarciaLnk:
`firewall-config` should be removed from includes, rather than added to excludes

fesdonomist:
You need to remove the .desktop file added in https://github.com/ublue-os/aurora/commit/198a96107c8fb1b8845fe83d84923683cb172b94

PR #557, opened 2025-06-05
feat: set Framework launcher icon for Framework devices

ledif:
This PR checks if we are running on a Framework laptop / desktop and sets the icon for the application launcher to the Framework icon. It also removes the GNOME-specific code that was copied directly from Bluefin.

I tested the script by manually running it but it would be helpful to test in a better way.

image

PR #560, opened 2025-06-06
fix: rework user setup hooks to use ublue-setup-services

ledif:
The `ublue-setup-services` package already provides the `/usr/libexec/ublue-user-setup` file, so the one in the Aurora repo is essentially dead code. This PR migrates the snippets that were in ublue-user-setup to hook files that the new script expects.

The only hooks that have not been migrated are the Framework specific snippets, which should be addressed in https://github.com/ublue-os/aurora/pull/557.

Haven't tested it yet, but it should be better than what we have now.


RealVishy (commented):
Is this a ptyxis theme that we ship?

RealVishy (approved):
Lgtm

PR #561, opened 2025-06-06
fix: remove Framework 13 (AMD) udev workarounds with latest BIOS

ledif:
This PR updates the Framework 13 (AMD) setup logic to stop applying the atkbd suspend udev workaround if the system is running BIOS version 3.09 or newer. Framework resolved the underlying issue last month in [BIOS 3.09](https://knowledgebase.frame.work/framework-laptop-13-bios-and-driver-releases-amd-ryzen-7040-series-r1rXGVL16) ("Fixed IRQ1 wake issues on Linux"), so the rule is no longer needed.

I manually invoked the script on my laptop and it successfully detected that I am running 3.09 and it removed the workaround.


inffy (approved):
LGTM

PR #563, opened 2025-06-06
chore: remove framework-kmod

inffy:
Seems this is not needed with kernel 6.14 anymore

ref:
https://community.frame.work/t/charge-levels-configurable-from-kde-userspace/69432/10


NiHaiden (dismissed):

inffy:
This might be better to do in main/akmods

PR #581, opened 2025-06-10
fix: jetbrains toolbox installation

inffy:
jetbrains-toolbox was updated and broke the ujust installation


inffy:
Should be good to go

Issue #579, opened 2025-06-10
Add Bazaar to Aurora


NiHaiden:
Add Bazaar to Aurora as the new app store. 

Add it once we have 

- [x]  Artwork 
- [x]  Defined our curated apps
- [x] figure out a logo

We will want to do something similar to Bluefin by hiding VS-Code and JetBrains Products so people don't install the wrong thing.

inffy:
> Will Bazaar sit alongside Discover? Discover is used for more than just Flatpak, so I'm hoping it won't be hidden or removed.

No decision yet on that part. For Aurora, discover doesn't really provide other than flatpak and lvfs support for the few oems that support it.

renner0e:
We also have to figure out how to disable discover search results in kickoff/krunner etc

inffy:
> We also have to figure out how to disable discover search results in kickoff/krunner etc

Yeah have to check with @ledif as he is working on the krunner integration.

renner0e:
How are we all feeling about replacing the discover icon on the panel and kickoff application launcher?

NiHaiden:
> How are we all feeling about replacing the discover icon on the panel and kickoff application launcher?

I would say we should do it and get people used to the new App Store. We can have Discover installed side-by-side with it for a transitionary period before we remove it.

inffy:
> > For Aurora, discover doesn't really provide other than flatpak and lvfs support for the few oems that support it.
> 
> Hmm, normally Discover also has Plasma Addons and Application Addons, but I see they're not present on Aurora for some reason. I guess most of the categories can also be found in the corresponding download button in system settings and other places. I guess technically not a loss if it wasn't already enabled, but maybe something to consider too (also Discover would handle updates of the installed things I think).

Those backends were removed a while ago

https://github.com/ublue-os/aurora/pull/425
https://github.com/ublue-os/aurora/pull/434

PR #585, opened 2025-06-10
feat: add bazaar and app list

dreamyukii:
add Bazaar and applications list #579


system_files/shared/usr/share/ublue-os/bazaar/blocklist.txt
```diff
@@ -0,0 +1,26 @@
+com.visualstudio.code
+com.visualstudio.code-oss
+com.vscodium.codium
```
etvt, line 3: Does this mean that I won't be able to install vscodium from Aurora's upcoming new official app store? What is the reasoning behind this please?

I currently have it installed and was using it with devcontainers (although there are some minor caveats).

What would be the alternatives? Using the installed vscode? I tried to avoid (at least starting, even if it's preinstalled) closed-source MS apps..

Thanks!

p5: For most users, the vscode flatpaks (both OSS and proprietary) don't work as they would expect.  This is due to the sandboxing applied to flatpaks, which make trivial things a lot more difficult.  So to reduce confusion, it's easier to hide it from the app stores and direct people to the tried, tested and supported vscode that's bundled with the image.

People can still install these applications with the CLI.

etvt: Thanks, CLI is good enough!

(I just hope they will improve over time the integration of these flatpak IDEs with the host to make it seamless for everybody.)

inffy:
I think  the applist/recommendations and the blocklist looks good. So we just need the art for the sections and afyer that it should be pretty good to go.

Also someone was tinkering with a qt frontend for this so we might see that also at some point.

dreamyukii:
> All of these changes were added in #677. Would you mind if this PR was closed?

yes

PR #665, opened 2025-07-03
docs: add building guide

renner0e:
this is kinda sorta a troubleshooting/building/testing guide, should this go into the docs, is this fine here?

https://github.com/renner0e/aurora/blob/building/BUILDING.md


inffy:
Might be better to build it into the docs and then just link it in the readme.

PR #695, opened 2025-07-15
feat(bazaar): Add Zen Browser to Browser section,  block sunshine

inffy:
Adds Zen Browser to browser section

Adds sunshine flatpak to blocklisted apps


renner0e (approved):
adding sunshine makes sense

PR #696, opened 2025-07-15
 chore: refactor Bazaar blocklist management 

ledif:
I think it makes sense to manage the Bazaar blocklist in packages.json so we can distinguish between `-dx` vs non-`dx` images. For example, we are hiding VS Code for both versions when we should really only be hiding VS Code for the `-dx` images.


ledif:
> I think generally the vscode flatpack is not advised to be used.
> 
> Generally dont like making the packages.json more messier as i think the longtime plan was getting rid of it.

Is that the case for all of the current Flatpaks in the blocklist (e.g., Android Studio, vscodium and JetBrains)? I thought that we are excluding these because they have system-level alternatives, not because of general "not recommended" reasons.

ledif:
Thanks for the input all. I'll close this issue but we should capture some of this discussion for the announcement when we move Bazaar out of beta in Aurora.

PR #709, opened 2025-07-18
fix(build): use GITHUB_TOKEN for curl when needed

ledif:
Based on https://github.com/rsturla/eternal-images/pull/391.


renner0e (approved):
confirmed that the local build still work

p5 (approved):
LGTM!

PR #723, opened 2025-07-22
feat: add dialout to dx-group ujust

inffy:
Adds the dialout group to the dx-group ujust for working with serial connections


ledif (approved):
LGTM. Seems to be based on https://github.com/ublue-os/bluefin/pull/2807.

NiHaiden (approved):

PR #728, opened 2025-07-23
feat(beta): add sudo-rs

renner0e:
following the lead on
https://github.com/ublue-os/bazzite/pull/2954

Beta only for now, Fedora should in my opinion push this change


inffy (approved):
Think we can put this in beta until we decide what we do with beta / possible testing or experimental branch3

PR #735, opened 2025-07-27
feat(beta): hide Discover and add ujust to re-enable

ledif:
[Screencast_20250727_192605.webm](https://github.com/user-attachments/assets/655b2ff1-5d7e-449a-bf1a-863e7705836a)


renner0e:
maybe it would be better if the recipe was hidden for now, non-beta users would get this as well

inffy (approved):

PR #736, opened 2025-07-28
fix(beta): update Bazaar banner alignment

ledif:
I forgot to commit a few changes in #734.


renner0e:
maybe it would be better if the recipe was hidden for now, non-beta users would get this as well

Issue #729, opened 2025-07-24
Migration pathway for `asus` and `surface` images


ledif:
Since we are deprecating `-asus` and `-surface` HWE images in the Fall, we have a few options that we could pursue to handle this when Fedora 43 is released.

1. We can simply remove the variants from the build matrix so that there will be no more updates for these images. Users would not receive any updates and may not even notice.
2. We can additionally remove the packages containing the words `asus` and `surface` from the registry, so that users who manually try to update will receive an error.
3. We can further ship a systemd one-shot service unit shortly before the release of Fedora 43 that will automatically rebase users on these images to the standard HWE images. This service can also notify the user what is happening.

Option 2 is how the Bazzite team recently handled removing their `bazzite-dx` images with `nvidia-open` because they were the same as the `nvidia` images. This option has a few drawbacks:
- The error that the user gets when updating is vague and unhelpful. It does not communicate that the image itself has been removed and users should rebase to something else.
- If a user never manually updates, they might not even notice that there will never be anymore updates for their image
- People are now posting in Discord asking for help because their `bazzite-dx` updates are not working.

I think we should instead pursue Option 3 for HWE images. It should preemptively fix issues that users will encounter once we remove the images and can help users better understand what is happening.

inffy:
Option 3 would be propably the best. Depcrecation of these images have been announced few times but its propably already been forgotten by the actual users, or some of them propably don't even know.

inffy:
I mean we haven't had ISOs for the HWE images for long time, so atleast no "new installs" have happened. If someone has rebased to them, then they usually know what they have done.

ghrc keeps the images for 90 days, we won't be remove those anyway, that is done automatically.

ledif:
This is not stale. It will be implemented before Fedora 43.

PR #744, opened 2025-07-30
feat(beta): add eol-rebaser and HWE migration config

ledif:
Implements option 3 from https://github.com/ublue-os/aurora/issues/729 using https://github.com/ledif/eol-rebaser.

I'm still in the process of testing.


inffy:
FYi, I migth remember wrong but...

I faintly remember that the plan, atleast at some point, was that we will remove the HWE images completely.

inffy:
> > I faintly remember that the plan, atleast at some point, was that we will remove the HWE images completely.
> 
> Quick question about it: would that be with the goal of integrating these HWE things into the main image, of waiting until they're somehow mainlined, or just removing and getting rid of the extras and whatever they supported?

Just total removal. surface/asus stuff would be dropped, there is not much more in the HWE images other extras than those few packages for surface/asus addition to the bazzite kernel

PR #762, opened 2025-08-01
feat: move to ublue-os devcontainer

inffy:
Picked from https://github.com/ublue-os/bluefin/pull/2903


RealVishy (dismissed):
Nice

.devcontainer
ledif, line 1: I think this should be named `.devcontainer.json` instead of just `.devcontainer`.

ledif (changes requested):

.devcontainer
inffy: true true

ledif (approved):
LGTM

Issue #773, opened 2025-08-03
[PROPOSAL] Change the default filesystem from btrfs to xfs


inffy:
We've been "suffering" from the current BTRFS kernel regression for over a month now and have had to pin stable to 6.14.11 (which is EOL).

There was some talk in Discord for Bazzite and Bluefin looking maybe moving to XFS. Bluefin-lts uses XFS as it comes from CentOS.

Didn't yet look too deep into it but atleast on the ISO configuration
https://github.com/ublue-os/aurora/blob/bcc3d5ee4e1e988a355304047e609dc093aab094/iso_files/configure_iso_anaconda.sh#L64

inffy:
Main differences seem to be:

```
XFS vs BTRFS trade-offs**:
   - **XFS**: Better performance for large files, mature and stable, good for high-performance workloads
   - **BTRFS**: Built-in compression, snapshots, copy-on-write, better space efficiency

3. **Feature differences**:
   - You'll lose BTRFS features like compression (`zstd:1`), snapshots, and subvolumes
   - XFS doesn't support filesystem shrinking (only growing)
 ```

bdols:
my 0.0002 cents: why not ext4 for the root and maybe /var and xfs for /home? ext4 is mature and stable and I read a few concerning comments on hacker news a while back about btrfs surrounding sporadic performance dips which dissuaded me from using it. XFS has a whole different repair/recovery tool set and process that I'm not sure how the boot process handles it as compared to ext4. I believe that repairing a XFS root partition needs to be ran on an unmounted filesystem meaning offline while booted into a rescue disk, i.e. sysrescuecd, where as e2fsck can repair a root filesystem mounted read-only.

renner0e:
To quote Jorge on discord:

> and also there is exactly 1 enterprise product using bootc and that uses xfs. That's important to me because those people are banks and stuff

I'd be +1 for this. If it means running a boring filesystem so we don't have to run an EOL kernel for that long ever again.

People who care about their filesystem can just choose another, plus we probably wouldn't have as much confused users who feel the need to have `/` snapshots.

castrojo:
Yeah it'd be worth it just to not have to answer the question about snapshots for root hahah.

Danathar:
If any filesystem should replace BTRFS in oder to keep features like compression and snapshots it would be ZFS. ZFS is arguably the most mature and stable file system out there. 

[...]

PR #813, opened 2025-08-10
feat: move new installs to xfs

inffy:
Moves the ISOs to use XFS.

This is currently totally untested though

https://github.com/ublue-os/aurora/issues/773


renner0e:
this is everything with default settings, automatic partitioning
image

inffy:
I think this is ready

iso_files/configure_iso_anaconda.sh
```diff
@@ -62,12 +62,10 @@ efi_dir = fedora
...
 
 [Storage]
-default_scheme = BTRFS
-btrfs_compression = zstd:1
+file_system_type = xfs
 default_partitioning =
     /     (min 1 GiB, max 70 GiB)
-    /home (min 500 MiB, free 50 GiB)
-    /var  (btrfs)
+    /var  (min 5 GiB, free 50 GiB)
```
tulilirockz, line 68: Lets change the minimums a bit here, `/var` is getting too cramped for me (i use rootful storage for a ton of stuff a lot). Also, lets bump /boot

inffy: fedora is bumping boot to 2gigs I think with F43. But where does the /boot size come as its not currenttly on the config

idnovic: https://github.com/rhinstaller/anaconda/pull/6678/files

idnovic: I do support increasing boot now. I do not want to reinstall in 6 months.

idnovic: What about keeping btrfs progs and not mergin / and /home? This would allow a user to choose to use btrfs for home if they want to and further separate system and user data.

inffy: This is not going to happen in this cycle anyway so doesnt really matter currently if ever

PR #839, opened 2025-08-16
chore: remove ramalama from image

inffy:
fedoras package of ramalama is old and updates slowly. Moving this to brew

Also adds ujust recipe to install ai tools

Required: https://github.com/ublue-os/packages/pull/827


NiHaiden:
LGTM

PR #848, opened 2025-08-19
fix(rollback-helper): include nvidia and other image variants

inffy:
rollback-helper doesn't show nvidia or other variants currently.

picked fomr https://github.com/ublue-os/bluefin/pull/3025/


inffy:
should be fixed now 

image

PR #886, opened 2025-08-25
chore(bazaar): remove org.gnome.Builder

ledif:
This probably shouldn't be here.


renner0e:
Why should it not be here?

PR #894, opened 2025-08-26
fix: Revert "feat(motd): add mention of bbrew to motd"

inffy:
Reverts ublue-os/aurora#867


renner0e (approved):

ledif (approved):
Since `bbrew` isn't installed by default, makes sense to not make it so prominent.

PR #906, opened 2025-08-30
chore: remove already installed packages from packages.json

renner0e:
These are all already in our base image
`quay.io/fedora-ostree-desktops/kinoite:42`

I did not touch any packages we add in main (that we maybe shouldn't have added in the first place), like tmux.

Tested by diffing the output of `rpm -qa | sort` in each container

draft for now, wait until next stable builds, just in case


packages.json
```diff
@@ -45,30 +37,24 @@
...
 				"ptyxis",
-				"pulseaudio-utils",
 				"python3-pip",
 				"rclone",
 				"restic",
 				"rsms-inter-fonts",
-				"samba",
-				"samba-dcerpc",
-				"samba-ldb-ldap-modules",
+				"samba-winbind",
```
ledif, line 50: Seems like you added a package here.

ledif (commented):
Could you provide the full command that you used to compare the list generated by `rpm -qa` and `packages.json`? I think it would be helpful that others can try the same thing as this list is bound to drift for various versions of Fedora.

renner0e:
> Could you provide the full command that you used to compare the list generated by `rpm -qa` and `packages.json`? I think it would be helpful that others can try the same thing as this list is bound to drift for various versions of Fedora.

oh yeah samba-winbind, technically not needed as it is pulled anyway by the other 2.

build from this branch

`just build aurora stable`
`just run`
inside the container
`rpm -qa | sort | fpaste`
download file on host

do this for main branch as well and with `just build aurora-dx stable`

`vimdiff pr906-vanilla.txt main-vanilla.txt`
`vimdiff pr906-dx.txt main-dx.txt`

also https://github.com/ublue-os/aurora/actions/runs/17353235535/job/49262583286#step:9:6219 and https://github.com/ublue-os/aurora/actions/runs/17353235535/job/49262583287#step:9:6219

PR #975, opened 2025-09-15
fix: changelogs better formatting for authors

inffy:
added some fixes for the authors section, specifially with names (spaces etc)

https://github.com/ublue-os/bazzite/pull/3221

still would  prefer to use github usernames and link the username to their profile but haven't figured out a way to make it work correctly


ledif (approved):
LGTM

PR #976, opened 2025-09-15
feat: add shortcut for offline docs

inffy:


inffy:
Not sure if this is  something we  want, or a good solution but couldn't really think anything else for now

Issue #990, opened 2025-09-19
Fix iso checksum file naming


renner0e:
We need to steal this
https://github.com/ublue-os/bluefin/pull/3236/files

PR #993, opened 2025-09-19
fix: generate ISO checksums after renaming

inffy:
closes https://github.com/ublue-os/aurora/issues/990

from https://github.com/ublue-os/bluefin/pull/3236


renner0e (approved):
works
image

PR #1033, opened 2025-10-01
chore: remove patched switcheroo

renner0e:
Wait for F43
https://src.fedoraproject.org/rpms/switcheroo-control


inffy:
Once ZFS updates the next build will be f43 as main and akmods are already updated

inffy (approved):

PR #1041, opened 2025-10-03
fix: offline docs away from panel

inffy:
Removes the offline docs icon from the KDE panel

@renner0e  mentioned that it doesn't also show on the Kickoff menus favorites part either, which is kind of  strange as its set up same as like Bazaars icon is..


ledif:
Is this still necessary, considering https://github.com/ublue-os/aurora/pull/1042 was merged?

ledif (approved):

PR #1044, opened 2025-10-03
feat(ci): add simple tests

renner0e:
the package test we already use in main

also added dx specific things


build_files/shared/build-base.sh
```diff
@@ -59,10 +59,13 @@ echo "::endgroup::"
...
+# Simple Tests
+/ctx/build_files/base/20-tests.sh
+
 # Clean Up
 echo "::group:: Cleanup"
 /ctx/build_files/shared/clean-stage.sh
 mkdir -p /var/tmp &&
     chmod -R 1777 /var/tmp
-ostree container commit
+bootc container lint --fatal-warnings || true
```
ledif:
```suggestion
bootc container lint
```
The `|| true` will hide actual errors. I think we should remove the fatal warnings flag since we don't want warnings to fail the build, but we still want actual errors to fail the build.

There are already a bunch of warnings that we should address eventually.
```
Lint warning: nonempty-boot: Found non-empty /boot:
  extlinux

Lint warning: sysusers: Found /etc/passwd entry without corresponding systemd sysusers.d:
  davfs2
  dhcpcd
Found /etc/group entry without corresponding systemd sysusers.d:
  ccache
  davfs2
  dhcpcd
  docker
  plugdev
  ...and 1 more

Lint warning: var-log: Found non-empty logfiles:
  /var/log/dnf5.log
  /var/log/dnf5.log.1

Lint warning: var-tmpfiles: Found content in /var missing systemd tmpfiles.d entries:
  d /var/cache/ccache 2770 root ccache - -
  d /var/cache/libdnf5 0755 root root - -
  d /var/cache/libvirt 0711 root root - -
  d /var/cache/libvirt/qemu 0750 root root - -
  d /var/cache/lxc 0755 root root - -
  ...and 56 more
Found non-directory/non-symlink files in /var:
  var/log/dnf5.log.1
  var/log/dnf5.log
  var/lib/containers/cache/blob-info-cache-v1.sqlite
  var/cache/ldconfig/aux-cache
  var/usrlocal/bin/overrides/swtpm
```

renner0e: we can expand the clean-stage script to address the /var ones

bluefin-lts does this:
https://github.com/ublue-os/bluefin-lts/blob/main/build_scripts/cleanup.sh#L15-L16

ledif: I think the `/var` warnings are benign, since they won't be carried over into the image. So we can deal with them later.

ledif (changes requested):
We should also add a call to `20-tests.sh` in `build_files/shared/build-dx.sh` as well. If we don't, the `-dx` builds would only have the tests in the middle of the build instead of the end.

renner0e:
> We should also add a call to `20-tests.sh` in `build_files/shared/build-dx.sh` as well. If we don't, the `-dx` builds would only have the tests in the middle of the build instead of the end.

add0ba92f8ae3d6520b1346638da961ba09f4c29

build_files/shared/build-dx.sh
```diff
@@ -48,6 +48,12 @@ echo "Variant=Developer Experience" >> /usr/share/kde-settings/kde-profile/defau
...
 /ctx/build_files/dx/09-cleanup-dx.sh
 
+# Simple Tests, shared with the base
+/ctx/build_files/base/20-tests.sh
+
+# dx specific tests
+/ctx/build_files/dx/10-tests-dx.sh
+
 # Clean Up
 echo "::group:: Cleanup"
```
ledif, line 58:
```suggestion
echo "::group:: Cleanup"
bootc container lint
```

If you look at the [Containerfile](https://github.com/ublue-os/aurora/blob/main/Containerfile), the dx stage is created after the base stage is built. So we still need a lint at the end of the dx stage.

ledif (approved):
Thanks for making these changes. LGTM!

PR #1050, opened 2025-10-05
add local speech support with Pied flatpak and speech-dispatcher

inffy:
picked from https://github.com/ublue-os/bluefin/pull/3320

local speech support to Aurora, enabling users to easily set up text-to-speech functionality with high-quality voices using Pied and speech-dispatcher.


renner0e:
Is this something that we can't achieve with flatpak then?

Issue #995, opened 2025-09-19
Bazaar crashes shortly after loading


darkview224:
### Describe the bug
I attempted to load Bazaar. After "retrieving remote entries" screen closes, the app crashes. This may occur instantly, or after a few seconds.

### Expected result
Bazaar would load and operate normally.

### Extra information
Previously (about a month ago), I'd experienced a similar bug on Discover, which dissuaded me from installing Aurora (I went with Ubuntu). I had loaded this ISO hoping the issue was resolved with Bazaar. At a minimum, it appears to be resolved with Discover. This is likely after a result of commits noted in similar issue 828: https://github.com/ublue-os/aurora/issues/828

After being prompted on Discord to file a report, this bug was replicated by a live ISO generated on 16-Sep.

I am aware that this processor is a bit under spec, but am hopeful that after the resolution of the Discover issue, it might also be resolved in Bazaar.

```
liveuser at aurora in ~ 
➜  kinfo
Operating System: Aurora 42
KDE Plasma Version: 6.4.4
KDE Frameworks Version: 6.17.0
Qt Version: 6.9.1
Kernel Version: 6.15.9-201.fc42.x86_64 (64-bit)
Graphics Platform: Wayland
Processors: 2 × Intel® Celeron® N4020 CPU @ 1.10GHz
Memory: 4 GiB of RAM (3.6 GiB usable)
Graphics Processor: Intel® UHD Graphics 600
```

```
Sep 19 22:27:19 systemd-coredump[3516]: Process 3197 (bazaar) of user 1000 term>
Sep 19 22:27:19 systemd[1]: Created slice system-drkonqi\x2dcoredump\x2dprocess>
Sep 19 22:27:19 systemd[1]: Created slice system-systemd\x2dcoredump.slice - Sl>
Sep 19 22:27:19 audit: BPF prog-id=105 op=LOAD
Sep 19 22:27:19 audit: BPF prog-id=106 op=LOAD
Sep 19 22:27:19 audit: BPF prog-id=107 op=LOAD
Sep 19 22:27:19 audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967>
Sep 19 22:27:19 systemd[1]: Started systemd-coredump@0-3516-0.service - Process>
Sep 19 22:27:19 audit: BPF prog-id=108 op=LOAD
Sep 19 22:27:19 audit: BPF prog-id=109 op=LOAD
[...]

Issue #948, opened 2025-09-08
investigate bazaar flatpak


renner0e:
to make it work you need all this:

make sure the host instance is not running

`pkill -f bazaar`

bazaar needs permission to `host-etc`

```
flatpak override --user io.github.kolunmi.Bazaar --filesystem=host-etc
```

- copy all the stuff we currently have in `/usr/share/ublue-os/bazaar/` to `/etc`
```
sudo cp -r /usr/share/ublue-os/bazaar /etc
```

- replace every instance of `/usr/share/ublue-os` in our `bazaar/config.yaml` with `/run/host/etc`
```
sudo sed -i 's|/usr/share/ublue-os/|/run/host/etc/|g' /etc/bazaar/config.yaml
```

This is how you actually run it:

```
flatpak run --command=bazaar io.github.kolunmi.Bazaar window --auto-service --extra-content-config /run/host/etc/bazaar/config.yaml
```

<img width="1885" height="949" alt="Image" src="https://github.com/user-attachments/assets/df083093-63c4-4784-91b5-68545685ab90" />

The krunner plugin somehow just works fine

<img width="2449" height="1276" alt="Image" src="https://github.com/user-attachments/assets/40345ada-da50-438c-851b-50f17cf88927" />

tulilirockz:
Just discovered that we can do overrides directly in `/var`, so we can create a `systemd-tmpfiles` entry that creates a symlink or maybe just copies our overrides directly into the system overrides so that they persist for every user. `/var/lib/flatpak/overrides` is the path

renner0e:
So thanks to the new preinstall stuff (which we will get in the next flatpak release soon™) the migration should not really be an issue

https://github.com/renner0e/diffuse/commit/c3665e6bb559414a38e1564fc7fdd729964924e7
https://github.com/renner0e/diffuse/commit/b745d655cde0a6dc902899ce5222802951be9c71

I am just gonna dump here how to unmark apps as preinstalled, so you can preinstall them, remove them to preinstall them again


`/var/lib/flatpak/repo/config` has this entry
```

```

renner0e:
https://github.com/ublue-os/aurora/pull/1159

Issue #845, opened 2025-08-18
duplicated entry for bazaar


renner0e:
<img width="863" height="325" alt="Image" src="https://github.com/user-attachments/assets/fc6f89d0-00b5-46e9-bec4-eb1a88190c3d" />

<img width="903" height="308" alt="Image" src="https://github.com/user-attachments/assets/6398b1b3-d59b-47ad-9435-46d1043ceab1" />

something super funky is going on, to reproduce:

- Make new user (new user has 2 entries, borked)
- `sudo bootc usroverlay && sudo rm /usr/share/applications/org.kde.discover.desktop` [(the symlink from discover to bazaar is gone)](https://github.com/ublue-os/aurora/blob/504b4fd34bb2057db6a8d2991d04630f83e762da/build_files/base/11-bazaar.sh#L17-L39)
- 2nd entry disappears

reboot, file will be restored as it's just a temporary overlay, so it should return to the borked state. right? right?

but the 2nd entry is gone for some reason

after that, removing `~/.config/plasma-org.kde.plasma.desktop-appletsrc` and then `systemctl --user restart plasma-plasmashell.service` makes it reappear again

this is all without having touched the [`ujust toggle-software-store`](https://github.com/ublue-os/aurora/blob/504b4fd34bb2057db6a8d2991d04630f83e762da/just/aurora-apps.just#L328)

PR #1076, opened 2025-10-11
feat: bazaar flatpak transition

renner0e:
same as: https://github.com/ublue-os/bluefin/pull/3352
resolves/adresses: https://github.com/ublue-os/aurora/issues/948 and https://github.com/ublue-os/aurora/issues/995

missing parts:
- https://github.com/ublue-os/packages/pull/957
- adapt ujust toggle-software-store for the flatpak if needed (figure it out with beta)
- investigate if we still need [this](https://github.com/ublue-os/aurora/blob/main/build_files/base/11-bazaar.sh#L34-L35)

We should have consensus first if we even want to do this like bluefin.
@ledif @NiHaiden @RealVishy @inffy
We have less stuff on the image, faster update cadence for bazaar, but people will be able to remove bazaar (for better or worse).

~~We should also in my opinion rip the band-aid off regarding the whole discover mess. I guess we have to because there will not be a bazaar desktop file in /usr anymore.~~

~~So don't do the whole [symlink mess](https://github.com/ublue-os/aurora/blob/main/build_files/base/11-bazaar.sh#L10-L25). But still disable discover by default and use bazaar.~~

~~We time this with F43 (beta) and we can tell people exactly what to click to remove the "blank discover icon" and pin bazaar instead, so no more trouble with [2 icons](https://github.com/ublue-os/aurora/issues/845), notifications from discover that have bazaar icon and all that trouble that comes with it and if people are bitching about bazaar then they can just remove it and suffer with discover.~~

Maybe [flatpak discover](https://github.com/renner0e/flathub/tree/discover) will be on flathub in the future...

### DIFF
--- a/image-versions.yml
+++ b/image-versions.yml
@@ -2,4 +2,4 @@ images:
   - name: kinoite-main
     image: ghcr.io/ublue-os/kinoite-main
     tag: latest
-    digest: sha256:231b0e427d3059df7adc6ed0d6fc01a250a4f73ba9c2437d7a12f2b1843f635f
+    digest: sha256:c82949a9e692f972190261269cc3d1bc4e6eb0b0fee1f5452afbca40a0ed0080
--- a/system_files/shared/usr/share/ublue-os/bazaar/config.yaml
+++ b/system_files/shared/usr/share/ublue-os/bazaar/config.yaml
@@ -189,6 +189,7 @@ sections:
       - com.feaneron.Boatswain
       - com.parsecgaming.parsec
       - net.davidotek.pupgui2
+      - org.endlessaccess.threadbare
 
   - title: "Utilities"
     subtitle: "Stuff you didn't need until now"


system_files/shared/usr/lib/tmpfiles.d/bazaar-flatpak.conf
ledif, line 1: Is there a way to do a drop-in for flatpak overrides?

 I worry with this solution, we are clobbering the path `/var/lib/flatpak/overrides/io.github.kolunmi.Bazaar` that the user may have already written to, if they manually included some overrides for Bazaar for whatever reason.

renner0e: not in /usr sadly, I tried it with /usr/share/factory on bluefin-lts but I could not make it work, at the end of the day the override needs to be in /var
[ublue-os/bluefin-lts@`5efec51` (#791)](https://github.com/ublue-os/bluefin-lts/pull/791/commits/5efec514c228678e3574561911391b53f5eaedc0)

PR #1093, opened 2025-10-16
chore: F43 building blocks

inffy:
Enables building of Fedora 43 betas.

Currently few small issues (neither one is blocker or something like that):

- Sunshine not available (no f43 packages available in the copr)
- openrazer-daemon not available

Also decided to rip off the scx-scheds copr+package


ledif (approved):
Nice!

Issue #391, opened 2025-04-27
Move more fonts to brew


inffy:
We could check which of our fonts that we install as an rpm could be moved to the homebrew brewfile.

Currently we have these fonts in our packages.json

`
firacode-nerd-fonts
jetbrains-mono-fonts-all
nerdfontssymbolsonly-nerd-fonts
rsms-inter-fonts

`

and in -dx

`
adobe-source-code-pro-fonts
cascadia-code-fonts
google-droid-sans-mono-fonts
google-go-mono-fonts
ibm-plex-mono-fonts
intel-one-mono-fonts
mozilla-fira-mono-fonts
powerline-fonts
google-noto-fonts-all
`

Currently in our brewfile for fonts we have:
`
cask "font-caskaydia-mono-nerd-font"
cask "font-droid-sans-mono-nerd-font"
cask "font-go-mono-nerd-font"
cask "font-blex-mono-nerd-font"
cask "font-open-dyslexic"
cask "font-sauce-code-pro-nerd-font"
cask "font-source-code-pro"
cask "font-ubuntu-nerd-font"
cask "font-fira-code-nerd-font"
cask "font-monaspace-nerd-font"
`
so atleast few that are overlapping.

Of the top of my head, just thinkin which are the fonts that need to be on the image are:
firacode, jetbrains(?),nerdfontssymbolsonly-nerd-fonts, powerline-fonts(?). 

I think the biggest "issue" is the google-noto one, its a huge package and fills the font list on users.

castrojo:
Yeah we have the same problem:

https://github.com/ublue-os/bluefin/issues/2425

Though IMO for the nerd/monospace fonts we should perhaps have those common across all DX variants, so perhaps we should centralize that? And then leave the desktop/proportional fonts to be more per image?

inffy:
> Though IMO for the nerd/monospace fonts we should perhaps have those common across all DX variants, so perhaps we should centralize that? And then leave the desktop/proportional fonts to be more per image?

That  sounds maybe the best way to have one centralized thing for nerd/monospace fonts.

castrojo:
https://github.com/ublue-os/packages/issues/393

m2Giles:
One thing to note for Aurora. KDE doesn't properly do font fallback at times so you need to use a NF.

PR #1103, opened 2025-10-17
chore: font cleanup

inffy:
Removed some of the fonts we are shippng, there was actually pretty many especially on the DX image.

I think these pretty much can be found from Brew.

Also changed the defaults so we are now aligned with Kinoite

fixes: https://github.com/ublue-os/aurora/issues/391

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -149,7 +149,7 @@
 	},
 	"42": {
 		"include": {
-			"all": [],
+			"all": ["uld"],
 			"dx": []
 		},
 		"exclude": {


packages.json
```diff
@@ -68,12 +66,10 @@
 				"zsh"
 			],
 			"dx": [
-				"adobe-source-code-pro-fonts",
 				"android-tools",
 				"bcc",
 				"bpftop",
 				"bpftrace",
-				"cascadia-code-fonts",
```
renner0e, line 76: what about vscode? I don't use it but this feels wrong. We should probably change it to something else for new users at least.
https://github.com/ublue-os/aurora/blob/main/system_files/dx/etc/skel/.config/Code/User/settings.json

inffy: aaargh

Umm well I don't know...Maybe put the default as DejaVu (although it should fallback anyway). Then we could add these fonts (most of them already are) in the fonts brewfile.

But we also need to add https://formulae.brew.sh/cask/font-symbols-only-nerd-font#default --> added this to the brew ai pr

renner0e:
> I was just thinking, should we wait until F43?

sure let's wait for F43 on that one, put it in the forum post as well and point people to the brew stuff

packages.json
renner0e: made an issue regarding this
https://github.com/ublue-os/aurora/issues/1127

Issue #1109, opened 2025-10-18
kickoff is busted and shows default kde stuff


renner0e:
these are all the PRs that may have caused it:

https://github.com/ublue-os/aurora/pull/976
https://github.com/ublue-os/aurora/pull/1041

~~our current iso from the website is still fine because it is old,~~ this only affects new user accounts that are created and future isos

I tested this by building an ISO with a the tagged releases to be 100% sure 

## good stable release before introduction of docs:
https://github.com/ublue-os/aurora/releases/tag/stable-20250921

<img width="981" height="834" alt="Image" src="https://github.com/user-attachments/assets/668df370-3dbb-487d-b573-de47816ace87" />

## good stable release after introduction of docs:
https://github.com/ublue-os/aurora/releases/tag/stable-20250928
<img width="1024" height="896" alt="Image" src="https://github.com/user-attachments/assets/1ae35bda-9638-413e-9590-4dcaa04f3dd8" />

## first busted release
https://github.com/ublue-os/aurora/releases/tag/stable-20251005

<img width="869" height="796" alt="Image" src="https://github.com/user-attachments/assets/836b0ca1-ec2b-4dad-84f1-38ace81a1230" />

PR #1110, opened 2025-10-18
fix: kickoff being weird

renner0e:
fixes: https://github.com/ublue-os/aurora/issues/1109
image


ledif (approved):
This seems much nicer. Looks good to me!

PR #1132, opened 2025-10-22
feat: add brewfile validation

inffy:
These are picked from Bluefin:
https://github.com/ublue-os/bluefin/pull/3450
https://github.com/ublue-os/bluefin/pull/3449

We already had the brewfiles on the image but i feel it's easier to find now that they are in the root brew folder and then copied to the correct folder.

### DIFF
--- a/.github/workflows/validate-brewfiles.yml
+++ b/.github/workflows/validate-brewfiles.yml
@@ -14,7 +14,8 @@ jobs:
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
 
       - name: Set up Homebrew
-        uses: Homebrew/actions/setup-homebrew@main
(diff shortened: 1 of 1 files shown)


.github/workflows/validate-brewfiles.yml
```diff
@@ -0,0 +1,39 @@
...
+      - "brew/**"
+jobs:
+  validate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
+
+      - name: Set up Homebrew
+        uses: Homebrew/actions/setup-homebrew@main
```
renner0e:
```suggestion
        uses: Homebrew/actions/setup-homebrew@8c478e7f5c2c2df9efba834a3568d4fe35cdb2a3

```
I think we should make codacy happy, they don't seem to be updating it super often so the spam is fine in my opinion

renner0e (approved):

PR #1168, opened 2025-10-25
fix: add nerdfont-symbols

inffy:
before:
image

after:

Screenshot_20251025_183030

### DIFF
--- a/image-versions.yml
+++ b/image-versions.yml
@@ -2,4 +2,4 @@ images:
   - name: kinoite-main
     image: ghcr.io/ublue-os/kinoite-main
     tag: latest
-    digest: sha256:0c20ddc06d20ae81e69844cce86f10edf8569f1bb9d724bbee9fd30a09881df7
+    digest: sha256:c8ef7f1911cad3592ed1eaac9bd14f4caee9202310a7362f03f18c8a3710411b


renner0e:
I mean I guess that works, I was thinking more download the file, unzip, move

inffy:
> I mean I guess that works, I was thinking more download the file, unzip, move

oh you meant during the build. Oh well I guess it doesn't matter in the end

build_files/base/05-override-install.sh
```diff
@@ -15,6 +15,13 @@ install -c -m 0755 /tmp/starship /usr/bin
 # shellcheck disable=SC2016
 echo 'eval "$(starship init bash)"' >> /etc/bashrc
 
+# Nerdfont symbols
+# to fix motd and prompt atleast temporarily
+ghcurl "https://github.com/ryanoasis/nerd-fonts/releases/latest/download/NerdFontsSymbolsOnly.zip" --retry 3 -o /tmp/nerdfontsymbols.zip
+unzip /tmp/nerdfontsymbols.zip -d /tmp
+mkdir -p /usr/share/fonts/nerd-fonts/NerdFontsSymbolsOnly/
+mv /tmp/*.ttf /usr/share/fonts/nerd-fonts/NerdFontsSymbolsOnly/
```
ledif:
```suggestion
mv /tmp/SymbolsNerdFont*.ttf /usr/share/fonts/nerd-fonts/NerdFontsSymbolsOnly/
```
Let's narrow this glob, just in case `/tmp` happens to have random fonts in it (unlikely, but technically possible).


image

inffy: Right, didnt even think about that

renner0e (approved):

PR #1180, opened 2025-10-27
chore: more repo cleanup

inffy:
Align with bluefin https://github.com/ublue-os/bluefin/pull/3477

Not sure what is happening with the Beta builds, although tomorrow is the release so those won't be available for long anyway

### DIFF
--- a/.github/workflows/build-image-stable.yml
+++ b/.github/workflows/build-image-stable.yml
@@ -13,8 +13,8 @@ on:
       - "system_files/etc/bazaar/config.yaml"
       - "**.Brewfile"
       - "just/*.just"
-  # schedule:
-  #  - cron: "50 5 * * 0" # 5:50 UTC sunday
+  schedule:
+    - cron: "15 1 * * 2" # Every Tuesday at 1:15 AM
   workflow_call:
   workflow_dispatch:
 
--- a/.github/workflows/generate-release.yml
+++ b/.github/workflows/generate-release.yml
@@ -53,7 +53,7 @@ jobs:
 
       - name: Create Release
         uses: softprops/action-gh-release@6da8fa9354ddfdc4aeace5fc48d7f679b5214090 # v2
-        if: contains(fromJson('["stable"]'), matrix.version) && (github.event.schedule == '50 5 * * 0' || contains(fromJson('["workflow_dispatch", "workflow_call"]'), github.event_name))
+        if: contains(fromJson('["stable"]'), matrix.version) && (github.event.schedule == '15 1 * * 2' || contains(fromJson('["workflow_dispatch", "workflow_call"]'), github.event_name))
         with:
           name: ${{ steps.generate-release-text.outputs.title }}
           tag_name: ${{ steps.generate-release-text.outputs.tag }}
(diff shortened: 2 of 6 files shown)


build_files/base/05-override-install.sh
```diff
@@ -25,10 +25,51 @@ mv /tmp/SymbolsNerdFont*.ttf /usr/share/fonts/nerd-fonts/NerdFontsSymbolsOnly/
...
+ln -sf /usr/share/backgrounds/aurora/aurora-wallpaper-7/contents/images/3840x2160.jxl /usr/share/backgrounds/default-dark.png
+ln -sf /usr/share/backgrounds/aurora/aurora.xml /usr/share/backgrounds/default.xml
+
+# /usr/share/sddm/themes/01-breeze-fedora/theme.conf uses default.jxl for the background
+# We are lying about the extension
+ln -sf /usr/share/backgrounds/default.png /usr/share/backgrounds/default.jxl
+ln -sf /usr/share/backgrounds/default-dark.png /usr/share/backgrounds/default-dark.jxl
+
+# Favorites for Panel
+sed -i '/<entry name="launchers" type="StringList">/,/<\/entry>/ s/<default>[^<]*<\/default>/<default>preferred:\/\/browser,applications:org.gnome.Ptyxis.desktop,applications:org.kde.discover.desktop,preferred:\/\/filemanager<\/default>/' /usr/share/plasma/plasmoids/org.kde.plasma.taskmanager/contents/config/main.xml
```
renner0e, line 46: what do you think about putting the visual changes in a separate file?
we can do this in a separate PR. I think we can make the way we set the default apps in the panel for the task manager less cursed without the sed horrible_name.xml

inffy: sure, if we have a better way than the sed horror, i am all for it

Containerfile
```diff
@@ -23,32 +23,18 @@ ARG KERNEL="6.14.4-200.fc41.x86_64"
...
-# Build, Clean-up, Commit
-RUN --mount=type=cache,dst=/var/cache/libdnf5 \
-    --mount=type=cache,dst=/var/cache/rpm-ostree \
-    --mount=type=bind,from=ctx,source=/,target=/ctx \
-    --mount=type=secret,id=GITHUB_TOKEN \
-    /ctx/build_files/shared/build-dx.sh
+# Makes `/opt` writeable by default
+# Needs to be here to make the main image build strict (no /opt there)
+# This is for downstream images/stuff like k0s
+RUN rm -rf /opt && ln -s /var/opt /opt
```
renner0e: This is useful. What do the other think about this? If this doesn't break downstream images then I'm fine with it

NiHaiden: I am fine with it too, as long as it doesnt break anything existing.

inffy: Thinking out loud here, does anyone think anything that could break because of this? I don't see anything on our end breaking and not  sure if this  would affect any custom images that use Aurora as a base. Atleast nothing should go unbootable because of this.

renner0e: this is documented in the image-template now, everyone who needs that can use it, we should make it a separate PR at least and test this imo.

PR #1192, opened 2025-10-28
feat: plasma 6.5 for latest

renner0e:
stable is not affected

### DIFF
--- a/build_files/base/10-beta.sh
+++ b/build_files/base/10-beta.sh
@@ -4,17 +4,18 @@ echo "::group:: ===$(basename "$0")==="
 
 set -eoux pipefail
 
-PLASMA_VERSION=$(rpm -q --qf %{VERSION} plasma-desktop)
-
-# We don't want to install the copr version if it's in fedora repos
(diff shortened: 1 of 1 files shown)


build_files/base/10-beta.sh
```diff
@@ -4,19 +4,20 @@ echo "::group:: ===$(basename "$0")==="
...
-  echo "Skipping installing 6.5 when we are already on 6.5"
-  exit 0
-fi
 
 if ! jq -e '.["image-tag"] | test("beta|latest")' /usr/share/ublue-os/image-info.json >/dev/null; then
     echo "We only run this on latest and beta"
     exit 0
 fi
 
+PLASMA_VERSION="6.5.1"
```
miku-orb, line 13: Was this meant to be hardcoded?

renner0e: > Was this meant to be hardcoded?

Oh shit, I accidentally committed this I was testing if it would work 😭 

It should be this
`PLASMA_VERSION=$(rpm -q --qf %{VERSION} plasma-desktop)`

PR #1193, opened 2025-10-28
chore(ISO): Test building the webui iso with :latest

inffy:
Just a test to see if we can build this on the latest stream to get the newer webui and anaconda packages (which should make the webUI work on KDE)


inffy:
> I am assuming we need to redo the whole branding anyway? Because webui and we have proper art + logos now

yes, these still use the old stuff

inffy:
meh, doesn't work. Its mixing stuff from :latest but then using bits from quya.io/fedora which is still 42 base

PR #1210, opened 2025-10-30
feat: add fw-control and openrazer for all

inffy:
~~Draft to see once they are back and can be merged~~

Now both are available and installable on F43 too


inffy:
OpenRazer seesm to be available now

inffy:
Tested and works

image

PR #1219, opened 2025-10-31
fix: uncomment beta images properly

renner0e:


inffy:
asdf, i made a poopoo. We don't actually need any of those comments on hte file. They should be actually removed as the action is disabled.

So lets just remove all the # from the first "chapter" and leave the file disabled

inffy (approved):

PR #1223, opened 2025-11-01
chore(ci): comment out the beta script

inffy:
Comments the Beta.sh script (currently not in use anyway) and removes the kde-sig copr stuff


renner0e (approved):
LESSS GOOOO !!!!!!11

Issue #992, opened 2025-09-19
Since Sept 17, plasmashell keeps crashing and drkonqi fails to get debugging information


muelsyse-az:
### Describe the bug

Ever since September 17, I've been getting plasmashell crashes everyday every few hours or after waking up on sleep. When Dr Konqi tries to get information for sending an automatic report, it fails to do so (showing an error) and when force closing Dr Konqi, it crashes. 

### What did you expect to happen?

I don't expect constant crashes and I also don't expect Dr Konqi to fail getting crash information

### Output of `bootc status`

```shell
● Booted image: ghcr.io/ublue-os/aurora-dx:stable-daily
        Digest: sha256:f2ebb1924dbab415b68c682f4da58db2477095a01f89049476ac6bb6d12895b9 (amd64)
       Version: 42.20250918.1 (2025-09-18T13:16:30Z)
```

### Output of `groups`

```shell
fatomicreturns wheel docker incus-admin libvirt plugdev
```

### Extra information or context

It's a little difficult to reproduce so I'll send screenshots on another date. I'll just attach the many crash details that I can find in the Crashed Processes Viewer. 

If it helps, I have two panels. The top panel has external widgets Kara, Window Title, Window Spacer Extended and Cat Walk. The bottom panel acts as a dock with standard widgets. Top one is not floating, bottom one is floating and dodges windows.

Operating System: Aurora 42
KDE Plasma Version: 6.4.4
KDE Frameworks Version: 6.17.0
Qt Version: 6.9.2
Kernel Version: 6.15.10-200.fc42.x86_64 (64-bit)
Graphics Platform: Wayland
Processors: 8 × Intel® Core™ i5-10210U CPU @ 1.60GHz
Memory: 16 GiB of RAM (15.4 GiB usable)
Graphics Processor: Intel® UHD Graphics
Manufacturer: Dell Inc.
Product Name: Latitude 3410

[Sept17_1.txt](https://github.com/user-attachments/files/22418528/Sept17_1.txt)
[Sept17_2.txt](https://github.com/user-attachments/files/22418532/Sept17_2.txt)
[Sept17_3.txt](https://github.com/user-attachments/files/22418530/Sept17_3.txt)
[Sept18_1.txt](https://github.com/user-attachments/files/22418534/Sept18_1.txt)
[Sept18_2.txt](https://github.com/user-attachments/files/22418533/Sept18_2.txt)
[...]

PR #1229, opened 2025-11-01
feat(test): KDE and QT version mismatch

renner0e:
works as intended, failed our F43 builds successfully
checking for kscreen is jank, I'm guess this is an edge case in packaging somewhere
https://github.com/ublue-os/aurora/actions/runs/18999949910/job/54265252563?pr=1229
image


miku-orb:
I would also suggest to check the version match of kwin as well, I've ran into issues with it crashing when dnf partially upgraded main plasma + attempt to downgrade kwin in #992 also resulted in crashes similarly

PR #1142, opened 2025-10-23
chore: refactor package installation and handling during build time

inffy:
Refactor the package installation and COPR stuff

Closes https://github.com/ublue-os/aurora/issues/1115

PR #1230, opened 2025-11-01
chore: remove leftover kvmfr helper script + vfio stuff

renner0e:
leftovers from https://github.com/ublue-os/aurora/pull/1142


inffy (approved):

GarciaLnk:
The `configure-vfio` just recipe is still very useful even without the kvmfr module, I don't know why it was removed altogether, removing the `kvmfr` option would've been enough.

PR #1243, opened 2025-11-03
chore: remove devpod form default apps

renner0e:
EOL runtime

https://github.com/flathub/sh.loft.devpod/pull/5


etvt:
I think the devpod-based development workflow was one of the selling points of Aurora/Bluefin. What do you recommend then instead?

Issue #1245, opened 2025-11-04
F43 webui tracking issue


renner0e:
- we need our branding

- figure out how to get webui installer working, currently needs firefox rpm, steal from bluefin https://github.com/ublue-os/bluefin/blob/main/iso_files/configure_iso_anaconda.sh
a shim with flatpak run sadly doesn't work

~~- we can make them smaller by blowing away documentation in `/usr/share/docs` (200MB) and rpm db in `/usr/share/rpm/` (100MB)~~

- would be nice if we had a separate icon for the installer, either scope with a wrench or something, just scope, leaving it is also fine or a hard drive/installer icon with scope on top 

- at a later point investigate plasma-setup (formerly kiss) 

surprisingly the installation worked out just fine

<img width="2541" height="1423" alt="Image" src="https://github.com/user-attachments/assets/a5d54ab7-b52f-44d1-b359-184b7aa39358" />

renner0e:
@inffy yup, keyboard layout just worked, in webui as in plasma as well. Everything worked as expected.

renner0e:
> Did this include the new biggeer boot partition yet?
it does
<img width="940" height="141" alt="Image" src="https://github.com/user-attachments/assets/66604221-6a57-4627-bff0-9d3567e67274" />

PR #1246, opened 2025-11-04
feat: make webui ISO more pleasant

renner0e:
https://github.com/ublue-os/aurora/issues/1245

### DIFF
--- a/.github/workflows/build-image-stable.yml
+++ b/.github/workflows/build-image-stable.yml
@@ -14,7 +14,7 @@ on:
       - "**.Brewfile"
       - "just/*.just"
   schedule:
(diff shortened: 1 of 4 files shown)


renner0e:
not sure if I like it, I guess changing the icon in plasma-welcome-center to the regular Aurora one would be fine, if it is possible somehow
image

inffy:
We can move the ISO stuff back to `stable` from latest as it will be updated to F43 soonish, right?

renner0e:
> We can move the ISO stuff back to `stable` from latest as it will be updated to F43 soonish, right?

yup, we need to revert that change, I want to clean up the history anyway before merging, a lot of useless crap in here

PR #1260, opened 2025-11-06
Kiss

renner0e:
no worky so far, you'll get dropped into sddm without any user configured whatsover


renner0e:
wait for F44 for this, we don't need to do much there.

PR #1293, opened 2025-11-12
fix: fix Justfile to work with coreOS 43 builds

inffy:
from our akmods https://github.com/ublue-os/akmods/pull/416

With Fedora 43, CoreOS switched to a Containerfile (podman build) solution for publishing their images. As a side effect they no longer include the 'ostree.linux' label on those images.

This causes our coreos kernel version lookup to query RPM within a given image to find the current kernel rather than inspecting the image label.

This should update the Justfile to handle this for stable.

ZFS is not yet available for 6.17 kernel, so the builds are broken for now

### DIFF
--- a/.github/workflows/build-image-beta.yml
+++ b/.github/workflows/build-image-beta.yml
@@ -10,6 +10,7 @@ on:
        - "**.md"
        - "flatpaks/**"
        - "system_files/etc/bazaar/**"
+       - "system_files/shared/etc/bazaar/**"
        - "**.Brewfile"
        - "just/*.just"
    workflow_call:
--- a/.github/workflows/build-image-latest-main.yml
+++ b/.github/workflows/build-image-latest-main.yml
@@ -10,7 +10,7 @@ on:
       - "iso_files/**"
       - "flatpaks/**"
       - ".github/workflows/*iso*"
-      - "system_files/etc/bazaar/**"
+      - "system_files/shared/etc/bazaar/**"
       - "**.Brewfile"
       - "just/*.just"
   workflow_call:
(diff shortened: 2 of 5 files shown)


renner0e:
> So do we release without ZFS and put a huge "red flag" to our release post for possible ZFS users, or hold off until ZFS release is out (however long that will take) [openzfs/zfs#17857](https://github.com/openzfs/zfs/pull/17857)

They would auto-update to the next version, so I think we shouldn't push any builds that would potentially break crucial components like a filesystem. How long does it usually take for them to make a release after this gets merged? Like the people that desperately want F43 could have just rebased to latest weeks ago. KDE is still the same anyway.

It just sucks for us because of added complexity for now at least.

Justfile
```diff
@@ -127,7 +127,14 @@ build $image="aurora" $tag="latest" $flavor="main" rechunk="0" ghcr="0" pipeline
 
     # Kernel Release/Pin
     if [[ -z "${kernel_pin:-}" ]]; then
-        kernel_release=$(skopeo inspect --retry-times 3 docker://ghcr.io/ublue-os/akmods:"${akmods_flavor}"-"${fedora_version}" | jq -r '.Labels["ostree.linux"]')
```
GarciaLnk, line 130: this doesn't need to be changed, akmods still adds the ostree-linux label

bsherman: @inffy, this is correct and I came to this PR to say the same.

Upstream Fedora CoreOS no longer publishes the `ostree.linux` label, as it was apparently a side-effect of the build process, not deliberate.

But within unversal blue we've treated it as authoritative and `akmods` explicitly publishes the label. If you want to change your approach for choosing kernels, that's certainly fine, but there is no forced change for Aurora (or Bluefin) builds which use the `akmods:coreos-stable` images.

PR #1339, opened 2025-11-21
chore: update bazaar config to new format

renner0e:
https://github.com/kolunmi/bazaar/commit/f71dea0a3583d0ad36e28b3b5f8ff5356f864bc3

We need to keep the old txt one for now so the old bazaar version keeps working.

Some users could be running the old bazaar versions for a while.
Delete the txt one when we can safely assume that nobody runs the old version of bazaar anymore.
 
The txt format one will not be parsed by the new version. This is a breaking change that will happen in the next bazaar release, so we can land this before the actual bazaar release happens.

~~I still need to verify that this actually works.~~

image

image

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,7 +1,7 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
-    "config:best-practices",
+    "config:best-practices"
   ],
 
   // Do not rebase when the default branch is updated
@@ -37,7 +37,10 @@
     {
       "automerge": true,
       "matchUpdateTypes": ["digest"],
-      "matchDepNames": ["ghcr.io/ublue-os/kinoite-main"],
+      "matchDepNames": [
+        "ghcr.io/ublue-os/kinoite-main",
+        "ghcr.io/get-aurora-dev/common"
+      ]
     }
   ]
 }
--- a/.github/workflows/generate-release.yml
+++ b/.github/workflows/generate-release.yml
@@ -30,7 +30,7 @@ jobs:
 
     steps:
       - name: Checkout last 500 commits (for <commits> to work)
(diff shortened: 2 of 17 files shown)


system_files/shared/etc/bazaar/blocklist.yaml
```diff
@@ -0,0 +1,35 @@
+blocklists:
```
ledif, line 2:
```suggestion
# For the rationale about what is included in this list, see https://github.com/ublue-os/aurora/issues/786
blocklists:
```

system_files/shared/etc/bazaar/config.yaml
```diff
@@ -43,6 +45,10 @@ css: |
     color: #2d2f9b;
   }
 
+  .rec-section app-tile-verified-check {
```
AlexanderVanhee: I’d replace all `app-tile-verified-check` styles with the snippet below. Some of the existing variations have poor contrast or look a bit off.

```css
app-tile-verified-check {
  color: var(--fg-color);
}
```

I would also remove all `app-tile-installed-indicator` styles for mostly the same reasons, but no style overwrites are needed there.

AlexanderVanhee:
I would also recommend you use `border-radius: 12px;` on both `.global-section` and `.global-section banner` as this is the standard border radius for "card" elements in Adwaita.

system_files/shared/etc/bazaar/config.yaml
renner0e: Thanks!
Should be resolved with
https://github.com/ublue-os/aurora/pull/1339/commits/456522e5b902269b5fc78b6eeb9c5d4bf66ad457
https://github.com/ublue-os/aurora/pull/1339/commits/7dd11b58c8a324ed6c94fb3dbee307c93d1dfff4

PR #1340, opened 2025-11-21
feat: versioned isos with date

renner0e:
needs some website changes + cloudlfare bucket

the checksum stuff works
image


inffy:
maybe the date should be without the dashes, so like 20251121. what do you think?

NiHaiden:
I think the way to solve this is gonna be to upload a seperate metadata json file like current-iso.json or something along those lines that can be parsed by the download picker and then serve the current iso using the provided link in there. This is probably the easiest way to integrate it with the website and add support for it.

Thoughts?

PR #1374, opened 2025-11-29
chore: refactor get-aurora-dev/common image declaration 

ledif:
I think it's cleaner to pull the sha and image ref out of the Containerfile and into the Justfile, similar to other auxiliary images like rechunker.


renner0e:
looks like the verification is working :)
image

Issue #1400, opened 2025-12-02
Remove akonadi/mariadb from Aurora


hal7df:
So, I discovered somewhat by chance that Aurora has a full-fledged copy of MariaDB baked into the image...I did some digging and found that the reason for this is that MariaDB is required by Akonadi, which is a part of the `kde-desktop` group in dnf (which I think is what gets installed to provide "minimal KDE").  Akonadi is the storage backend for the KDE PIM suite (Kontact, KMail, KOrganizer, etc.), which is not installed in the base Aurora load; anyone using these programs are likely to [install them as a Flatpak](https://flathub.org/en/apps/org.kde.kontact) which [uses its own instance of akonadi](https://github.com/flathub/org.kde.kontact/blob/b67816b11ae3d7940032965b926ac7265794c818/org.kde.kontact.json#L211).

All told, Akonadi + MariaDB are taking up about ~100MiB of space when unpacked. Thus, my question: should Aurora (or, realistically, `kinoite-main`) just remove [these packages](https://github.com/hal7df/borealos/pull/24/files#diff-c508188958b68ce78976e33b66f47762f45f4a7bc134c5781cc38ee9ee81589a) since it seems Akonadi is basically dead weight in these images? Maybe someone is relying on the builtin mariadb package but honestly it seems better practice to pull in MariaDB either as a container image or via brew.

PR #1402, opened 2025-12-02
chore: free space - remove akonadi+mariadb

inffy:
closes https://github.com/ublue-os/aurora/issues/1400

### DIFF
--- a/build_files/base/04-packages.sh
+++ b/build_files/base/04-packages.sh
@@ -57,6 +57,7 @@ FEDORA_PACKAGES=(
     sssd-ipa
     sssd-krb5
     tmux
+    uld
     virtualbox-guest-additions
     wireguard-tools
     wl-clipboard


build_files/base/04-packages.sh
```diff
@@ -142,6 +144,13 @@ EXCLUDED_PACKAGES=(
     khelpcenter
     krfb
     krfb-libs
+    mariadb
+    mariadb-backup
+    mariadb-common
+    mariadb-cracklib-password-check
+    mariadb-errmsg
+    mariadb-gssapi-server
+    mariadb-server
```
renner0e:
```suggestion
    mariadb
    mariadb-common
    mariadb-errmsg
```
this is enough, we don't need to be super verbose here as these are automatically removed

renner0e (dismissed):
KDE Linux doesn't ship akonadi I guess this is fine to remove then.

renner0e (dismissed):

Issue #1404, opened 2025-12-02
Package uld missing (regression from Fedora 42)


miku-orb:
### Describe the package

Originally added in #322, after update to Fedora 43 base image, the Samsung printer doesn't work due to error: `File "/usr/lib/cups/filter/rastertospl" not available: No such file or directory`

Unsurprisingly why it's not included right now, since it's being excluded:
https://github.com/ublue-os/aurora/blob/cbe6fcc39ca361fc9ad77d606411b813eac8a473/build_files/base/04-packages.sh#L66-L78

I'm not sure what was the original source of the package, but the upstream has these instructions: https://negativo17.org/samsung-unified-linux-driver-printers-scanners/

And indeed the package is at version `1.00.39.12-3.fc43`. So perhaps it was withheld only temporarily back then due to lack of fc43 package? But it's there now, at least in that repo.

PR #1405, opened 2025-12-02
feat: add back uld samsung/hp printer driver

renner0e:
fixes: https://github.com/ublue-os/aurora/issues/1404 This is a proprietary driver from negativo repo.
We had this in the past and this is hardware enablement so I think it's fine to include (again).


inffy (approved):
this is fine, there wasn't f43 package previously as I think fedora dropped it or something. didn't know that t here was one in negativo

PR #1472, opened 2025-12-16
feat: move back image specific files/hacks/tech debt

renner0e:
partly undoes: https://github.com/ublue-os/aurora/pull/1421

The rationale behind this change is that we don't introduce tech debt to other aurora variants with this.

Please someone look over the files from common and add suggestions what would fit the criteria to be here.

### DIFF
--- a/build_files/base/19-initramfs.sh
+++ b/build_files/base/19-initramfs.sh
@@ -8,7 +8,7 @@ KERNEL_VERSION=$(rpm -q --queryformat="%{evr}.%{arch}" kernel-core)
 
 # Ensure Initramfs is generated
 export DRACUT_NO_XATTR=1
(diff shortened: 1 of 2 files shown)


system_files/shared/etc/rpm-ostreed.conf
```diff
@@ -0,0 +1,17 @@
+# Entries in this file show the compile time defaults.
+# You can change settings by editing this file.
+# For option meanings, see rpm-ostreed.conf(5).
+
+[Daemon]
+AutomaticUpdatePolicy=stage
+
+##########
+# NOTE: This will be set to true by default in Spring 2025
```
ledif:
```suggestion
# NOTE: There are plans to set to true by default in the future
```

system_files/shared/usr/lib/dracut/dracut.conf.d/10-compression.conf
ledif, line 1: This can be removed based on https://github.com/get-aurora-dev/common/pull/48#discussion_r2627422501

PR #1476, opened 2025-12-17
feat: rechunk with rpm-ostree

renner0e:
The legacy rechunker is basically unmaintained, this is a much safer
option to reduce our (delta) image size.

https://coreos.github.io/rpm-ostree/build-chunked-oci/

We can further reduce the layer diff by utilizing --previous-build in
the next rpm-ostree release. It still has a couple problems in the
current release in combination with non-rpm layers.

https://github.com/coreos/chunkah is very early in development, don't
bother for now.

### DIFF
--- a/build_files/base/17-cleanup.sh
+++ b/build_files/base/17-cleanup.sh
@@ -32,7 +32,7 @@ systemctl enable rechunker-group-fix.service
 systemctl enable flatpak-preinstall.service
 
 # disable sunshine service
-systemctl --global disable sunshine.service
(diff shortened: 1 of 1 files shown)


renner0e:
if we merge this then we should do this after the weeklies published so we get testing from our most technical users + we should make a blog post about this change only, no other topics, if something blows up we have a central place to do damage control of it, if it ever comes to that.

inffy:
Noted.

Also tested switching to image with this and nothing blew up. Also rollback was succesfull

PR #1504, opened 2025-12-20
feat: add containerd to base

inffy:
To be used with brew docker.

Step 1 to remove -dx


inffy:
Any thoughts on this?

PR #1534, opened 2025-12-27
fix(ci): use intermediate environment variables

renner0e:
mitigates script injection attacks
https://docs.github.com/en/actions/concepts/security/script-injections

Only doing the main image build for now to make this not a nightmare to review

2 ACKs would be nice :)


NiHaiden:
I read through your changes (and the docs) and it seems like it won't blow up, merging it.

PR #1604, opened 2026-01-07
chore(flatpak,iso): get flatpaks from common

inffy:
Needs: https://github.com/get-aurora-dev/common/pull/97

### DIFF
--- a/image-versions.yml
+++ b/image-versions.yml
@@ -2,7 +2,7 @@ images:
   - name: kinoite-main
     image: ghcr.io/ublue-os/kinoite-main
     tag: latest
-    digest: sha256:00f20c5ccc31e6c9861eeb73e9de8858329c843cde301d1970e251aee245ece0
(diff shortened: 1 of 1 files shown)


.github/workflows/reusable-build-iso-anaconda-webui.yml
```diff
@@ -73,12 +66,21 @@ jobs:
           echo "artifact_format="$image_name-${{ matrix.image_version }}-webui-$(uname -m)"" >> "${GITHUB_OUTPUT}"
           echo "kargs=$KARGS" >> "${GITHUB_OUTPUT}"
 
+      - name: Generate titanoboa-compatible file list
+        id: flatpak_list
+        run: |
+          set -eoux pipefail
+          FILE_LIST="$(mktemp)"
+          git clone https://github.com/get-aurora-dev/common.git common
```
tulilirockz: Using the checkout action is probably better than doing this tbh

inffy: But we would still need to parse the Brewfile right?

PR #1615, opened 2026-01-10
feat: move to upstream base images

renner0e:
Instead of using the ublue-os/kinoite-main base image we would be using
the ones built by Timothée over at https://gitlab.com/fedora/ostree/ci-test

This is a good path forward for Aurora because:

- we no longer have to undo things we inherit from ublue-os/main
- less overall dependencies
- we can make more conscious decisions about what goes into the image
- it's easier now to build images on top of upcoming Fedora versions
  (even KDE nightly ones) or to base on top of potential future images
  that are more aligned with our goals and needs

This would come at the cost of maintaining most of what main did
previously ourselves.

This PR is not supposed to clean stuff up, introduce as few changes as possible to what we currently have.

## Tests

- [x] Fedora flatpak is disabled, do we actually need the additional services from main? not a problem, we are not bringing that service in anyway because of our own flatpak version.
- [x] renovate, [seems to work](https://github.com/ublue-os/renovate-sandbox/pull/65/changes)
- [x] [only minimal changes are introduced](https://gist.github.com/renner0e/ce7d0eca98a3cc4a5864a60252b3b10e) (aurora:latest), seems fine to me, nothing that would break anything
- [x] all the coprs/repos disabled?
- [x] generally doesn't blow up due to the secure boot handling change (no ublue-os-akmods-addons). Booted for me with secure boot enabled so I guess it's fine.
- [ ] nvidia doesn't blow up for whatever reason (I don't have the hardware)

### Merge after weeklies


inffy:
So this changes our flow a little bit right (correct me if I am propably wrong lol)?

Current model:
akmods run with cron timer -> that kicks automatic ublue-os/main rebuild -> our renovate picks up the changed kinoite-main and updates our images.

What time does timothees builds run? Was just wondering that if they run before our scheduled akmods runs then we would be behind "one day" or something like that. Doesn't really affect that much, just wondering out loud.

renner0e:
> AT that point akmods is already built, so basically we don't need to renovate based on akmods(?).

I mean we could but we currently don't as well.
We don't explicitly rebuild stable or something when coreos releases, we just happen to get it on the next run.

PR #1657, opened 2026-01-19
fix: logos package conflict on downstream images

renner0e:
Not having *-logos leads to conclicts and will either break package
layering (file already exists) or completely replace them if used in
downstream images if a single package like anaconda requires them.

To comply with the licensing of the fedora-logos we have to swap it out
for the generic-logos.

"This package and its content may not be distributed with anything but
unmodified packages from Fedora Project. It can be used in a Fedora Spin,
but not in a Fedora Remix. If necessary, this package can be replaced by
the more liberally licensed generic-logos package."

By swapping it out with the generic-logos we can get away with not
making a dummy package, allowing these "problematic" packages to be installed
on downstream images and keep using OCI/non-RPM for our branding.

Fixes: https://github.com/ublue-os/bluefin/issues/4048

### DIFF
--- a/build_files/shared/build.sh
+++ b/build_files/shared/build.sh
@@ -4,8 +4,12 @@ set -eoux pipefail
 
 echo "::group:: Copy Files"
 
-# We need to remove this package here because lots of files we add from `{projectbluefin,get-aurora-dev}/common` override the rpm files and they also go away when you do `dnf remove`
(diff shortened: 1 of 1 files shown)


renner0e (commented):
There is probably a more elegant way on the removal side of things to this.
We definitely need to diff the files in this PR to make sure we are not removing unrelated files that are not part of the -logos package

Issue #1694, opened 2026-02-01
refactor fedora version detection


renner0e:
this currently relies on hacks which broke our beta builds:

https://github.com/ublue-os/aurora/actions/runs/21554415733/job/62108286874?pr=1692

My suggestion is to set the fedora version manually for each tag like stable, latest, beta and then check if akmods are available for that image flavor (coreos-stable, main) instead of ublue-os/main which we don't rely on anymore.

We use manifest.json as a way to not query the registry with skopeo inspect every time for local builds to detect the fedora version

PR #1696, opened 2026-02-01
fix(ci): refactor fedora version logic

inffy:
Closes https://github.com/ublue-os/aurora/issues/1694

Not sure what would be the best way to do this and if this is what was suggested.

So this requires us to always bump the tag versions when new Fedora releases come.


renner0e (approved):
I think this is fine, the json file just to store the fedora version is a little overkill in my opinion, but this is also used in the current rechunker logic, let's do that in https://github.com/ublue-os/aurora/pull/1476 as that whole section gets removed there anyway

PR #1771, opened 2026-02-15
refactor: reorder build scripts

renner0e:
Many RUN statements are very costly in terms of disk IO as every single
one commits layer cache for no benefit. But this retains most benefits
of the split in terms of disabling networking on the scripts that don't
need it.

There were also some leftover scripts that were running one command or
beta.sh which we don't use at all anymore. If there is a need to do
something on the beta only we should probably integrate it in the
other scripts.

Due to moving image-info.sh to the end we can drop the copr.vendor.conf
as we are still fedora when installing packages.

The disablement of openh264 repo and ublue akmods copr in dx.sh we can drop as
this is already handled by cleanup.sh

This is how https://github.com/ublue-os/aurora/pull/1670 should've been
done from the start.


renner0e:
needs to be tested for parity as the whole dx specific package stuff moved before akmods

PR #1798, opened 2026-02-21
feat: add more tesseract-langpack to the image

dreamyukii:
add all tesseract-langpack to the image


renner0e:
I kinda dislike this :/, but I don't have a better solution. 

I feel like it's not a very good use of space to add the entirety of 123 language packs for everyone which results in additional 323MiB, maybe a compromise of the 10 or so most spoken languages would be fitting?

PR #1816, opened 2026-02-26
chore(ci): store SBOMs as OCI artifact

renner0e:
Our current approach by uploading our rather large SBOMs to rekor doesn't work anymore for unknown reasons to me and is breaking our changelog generation in turn.

The compressed SBOMs are now a couple MiB big and stored on ghcr like everything else which should make things easier for us.

This means we have to adjust our changelog script again :(


m2Giles:
Other item, we still need to sign the attached image as well.

renner0e:
> Other item, we still need to sign the attached image as well.

https://github.com/renner0e/aurora/actions/runs/22458089591/job/65044635710
https://github.com/ublue-os/aurora/pull/1816/changes/f2d04fc129e63fa6b1af881ac353029b4ca0416e

inffy (approved):

PR #1847, opened 2026-03-06
chore: get nvidia rpms from akmods, remove i686

renner0e:
This means we don't have to download the nvidia stuff anymore as it
comes bundled with the nvidia akmods and should be a little more
reliable in general.

We don't need 32-bit libraries on the host, people should use Steam and
friends from Flathub and distrobox, this makes the nvidia images
roughly 800MiB lighter.

needs: https://github.com/ublue-os/akmods/pull/457


renner0e:
fine to merge, next akmods build will also install that one noarch package from akmods instead of repos

inffy (approved):

PR #1851, opened 2026-03-07
chore: remove starship

### LINKED ISSUE #1742: Remove starship from the image and encourage brew package instead
Remove the preinstalled starship package and promote the brew one instead. Announce via the blog and provide some sort of migration guide if any is necessary.

renner0e:
Starship can be installed with homebrew, for multi user systems you can either setup the shared linuxbrew group or download the binary from the github releases and drop it in /usr/local/bin as that is wriable or ~/.local/bin.

fixes: https://github.com/ublue-os/aurora/issues/1742


inffy (approved):

T045T:
It's a bit unfortunate that get-aurora-dev/common still has a file that tries to configure Starship: https://github.com/get-aurora-dev/common/blob/main/system_files/shared/etc/profile.d/90-aurora-starship.sh

Now every time I open a shell, that fails and complains.

I opened an issue on get-aurora-dev/common, since that is what's currently causing the undesired behavior: https://github.com/get-aurora-dev/common/issues/173

PR #1865, opened 2026-03-10
chore: improve ssf scorecard

inffy:
small permission tweaks to our workflows

### DIFF
--- a/.github/workflows/reusable-build.yml
+++ b/.github/workflows/reusable-build.yml
@@ -263,7 +263,7 @@ jobs:
           name: ${{ env.IMAGE_NAME }}.oci
           path: ${{ env.IMAGE_NAME }}_build
           if-no-files-found: error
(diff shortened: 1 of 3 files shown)


.github/workflows/validate-renovate.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       - name: Install dependencies
         shell: bash
         env:
-          RENOVATE_VERSION: latest
+          RENOVATE_VERSION: "43.61.0"
```
renner0e: Renovate made like 5 releases in the past 24h only, we better come up with something that wouldn't spam renovate bump PRs, this is gonna be very annoying

inffy: just noticed that they are doing releases all the time. So i guess the "latest" is the better option here

PR #1866, opened 2026-03-10
feat: add OpenSSF scorecard regression check

inffy:
Adds a small regression check for OpenSSF scorecard.

Uses bootc-dev/actions/openssf-scorecard action to do the regression check

Regression check runs with new PRs and fails if the baseline score drops

### DIFF
--- a/.github/workflows/reusable-build.yml
+++ b/.github/workflows/reusable-build.yml
@@ -263,7 +263,7 @@ jobs:
           name: ${{ env.IMAGE_NAME }}.oci
           path: ${{ env.IMAGE_NAME }}_build
           if-no-files-found: error
-          retention-days: 7
+          retention-days: 1
 
       - name: PR Testing Instructions
         if: github.event_name == 'pull_request'


.github/workflows/scorecard.yml
```diff
@@ -76,3 +78,22 @@ jobs:
...
+    runs-on: ubuntu-slim
+    if: github.event_name == 'pull_request'
+    permissions:
+      contents: read
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+          fetch-depth: 0
```
renner0e, line 93: > Only a single commit is fetched by default, for the ref/SHA that triggered the workflow. Set fetch-depth: 0 to fetch all history for all branches and tags. Refer [here](https://docs.github.com/actions/using-workflows/events-that-trigger-workflows) to learn which commit $GITHUB_SHA points to for different events.

so I'm understanding this right. if this only runs on a PR, why fetch all tags and branches?

inffy: as far as i understoot, it is needed so git can compute the merge base between the base branch and the PR head

renner0e (approved):

PR #1872, opened 2026-03-11
chore(ci): add attestation

renner0e:
This proves how and where aurora was built

https://github.com/renner0e/aurora/attestations


.github/workflows/reusable-build.yml
inffy, line 38:
```suggestion
      id-token: write
      attestations: write
```
should this have also permissions for th attestations?

PR #1879, opened 2026-03-12
chore(beta): make F44 build

renner0e:
feel free to further work on this to get this green!

### DIFF
--- a/.github/workflows/build-image-beta.yml
+++ b/.github/workflows/build-image-beta.yml
@@ -1,6 +1,6 @@
 name: Beta Images
 on:
-   merge_group:
+   #merge_group:
    pull_request:
      branches:
        - beta


inffy:
Should we maybe disable nvidia builds for now? Don't know if we have a way to exclude the libnvidia-container-tools package from the nvidia installation script? @renner0e

build_files/base/01-packages.sh
```diff
@@ -264,17 +263,11 @@ rm -rf /usr/share/plasma/look-and-feel/org.fedoraproject.fedora.desktop/
 
 # https://github.com/ublue-os/bazzite/issues/1400
 # TODO: test if we still need this when upgrading firmware with fwupd
+dnf5 -y copr enable "ublue-os/staging"
 dnf5 -y swap \
   --repo=copr:copr.fedorainfracloud.org:ublue-os:staging \
   fwupd fwupd
-
-# TODO: remove me on next flatpak release when preinstall landed in Fedora
```
renner0e, line 272: I forgot to point this out but we still need this because of our patch that is not yet in the tagged release

Issue #1704, opened 2026-02-02
make sure plasma-setup has branding


renner0e:
upstream issue:
https://invent.kde.org/plasma/plasma-setup/-/issues/72

we will have to patch plasma-setup which will be very annoying ~~or we could perhaps make a symlink from the fedora them to ours we should look into that~~

inffy:
https://github.com/ublue-os/packages/pull/1191

Issue #1115, opened 2025-10-19
get rid of packages.json


renner0e:
right now the problem is that:

- we enable a bunch of coprs
- we install all of our stuff in one big step, some stuff from fedora, some stuff from some copr, while copr takes priority over fedora repos.
- if a malicious copr adds a package like wl-clipboard, which we get from fedora and should only get from fedora, then we will ship a malicious wl-clipboard from a copr that we use for something else entirely
- we disable them at the end after everything and would not realize we shipped a malicious copy of wl-clipboard

so we would roughly do it like this for every copr package

```sh
# install all the stuff from the fedora repos
# enable copr
# dnf swap chungus chungus-patched-from-copr
# disable copr
# install stuff from fedora again if we have to for some odd reason
```

inffy:
With F43 I think we can generally remove Terra I think.

If we want something from there, we can do the enable / disable trickery and then just install packages specific from terra.

Or am I reading something wrong regarding the pkg list from renner couple comments up

PR #1916, opened 2026-03-16
chore(beta): plasma-setup branding

renner0e:
fixes: https://github.com/ublue-os/aurora/issues/1704

Needs: https://github.com/ublue-os/packages/pull/1191


inffy (approved):

build_files/base/01-packages.sh
```diff
@@ -286,7 +284,17 @@ dnf5 -y --repo=copr:copr.fedorainfracloud.org:ublue-os:flatpak-test install flat
 #    dnf5 upgrade --refresh --advisory=FEDORA-2024-dd2e9fb225
 #fi
 
-dnf -y install plasma-firewall-$(rpm -q --qf "%{VERSION}" plasma-desktop)
+# https://invent.kde.org/plasma/plasma-setup/-/issues/72
+dnf -y copr enable ublue-os/staging
+dnf -y copr disable ublue-os/staging
```
ledif, line 289: I'm a bit confused what these two lines are doing -- enabling and then immediately disabling?

renner0e: so only that repo is enabled for that transaction, we have been doing that this way since the big cleanup last cycle, we are just using a wrapper for most things, which is not possible here
https://github.com/ublue-os/aurora/issues/1115

PR #1928, opened 2026-03-19
fix(beta): replace konsole with ptyxis in plasma

renner0e:


ledif (dismissed):
LGTM but I think this PR should be titled "fix(beta): replace konsole with ptyxis"

inffy (approved):

PR #1978, opened 2026-03-31
feat(ci): consume previous build

renner0e:
Currently we compose a fresh rechunked image every build. This is not
ideal for image layer difference reduction and thereforce update sizes
as we do not take the rechunking plan/package layout from the previous
build (that users will likely have downloaded) into account.

When we have a new release of rpm-ostree we can replace most of the
pulling logic here with the new `--previous-build` flag.

Over time the layer plan due to reuse of the previous build's rechunking
plan might become undesirable/inefficient or due to issues in rpm-ostree
itself might force us to reset the layer plan where we do not consume
our previous build.

We need to check if the previous build was actually made with
rpm-ostree, otherwise this isn't going to do anything, implemented like
how upstream checks for this, see the comment.

image

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -293,8 +293,14 @@ rechunk $image="aurora" $tag="latest" $flavor="main" ghcr="0" pipeline="0" previ
       # TODO: Redo everything here with --previous-build in rpm-ostree 2026.1+
       # so we don't have to pull an old image + rename it
     if [[ "{{ previous_build }}" == "1" ]]; then
-      PREVIOUS_IMAGE=ghcr.io/{{ repo_organization }}/"${image_name}":"${DEFAULT_TAG}" 
-      ${SUDOIF} ${PODMAN} pull ${PREVIOUS_IMAGE}
+      PREVIOUS_IMAGE=ghcr.io/{{ repo_organization }}/"${image_name}":"${DEFAULT_TAG}"
+
+      # https://github.com/coreos/rpm-ostree/blob/7e2f2065a4aa4d5965b4537bb7d74e0b2898650e/rust/src/compose.rs#L522-L529
+      if skopeo inspect docker://"${PREVIOUS_IMAGE}" | jq -e '.LayersData[1:] | all(.Annotations?["ostree.components"]?)'; then
+        ${SUDOIF} ${PODMAN} pull ${PREVIOUS_IMAGE}
+      else
+        echo "${PREVIOUS_IMAGE} doesn't exist. Making a fresh layer Plan instead."
+      fi
     fi
 
     if [[ "{{ ghcr }}" == "1" ]]; then


.github/workflows/build-image-beta.yml
```diff
@@ -17,6 +17,14 @@ on:
...
    workflow_call:
    workflow_dispatch:
+    inputs:
+      previous_build:
+        description: 'Set to 0 to manually Clear Rechunking Plan'
+        type: choice
+        default: '1'
+        options:
+          - '0'
+          - '1'
```
renner0e, line 27: this is horrible, I have no idea how to make this so we only set this in the reusable workflow once and not in twelve other places and have it actually show up in the github ui on workflow dispatch

Justfile
```diff
@@ -292,10 +292,20 @@ rechunk $image="aurora" $tag="latest" $flavor="main" ghcr="0" pipeline="0":
       {{ just }} load-rootful "${image}" "${tag}" "${flavor}"
     fi
 
-    # In CI this will replace the unrechunked image
+      # TODO: Redo everything here with --previous-build in rpm-ostree 2026.1+
+      # so we don't have to pull an old image + rename it
+    if [[ "{{ previous_build }}" == "1" ]]; then
+      PREVIOUS_IMAGE=ghcr.io/{{ repo_organization }}/"${image_name}":"${DEFAULT_TAG}"
```
renner0e: mental note that I still have to implement something that checks if
- the image exists on the registry at all
- is actually made with rpm-ostree

before pulling

I have written something in the past that looked like this

```yaml
      # this tries to make delta updates smaller by pulling
      # the last image we published under this tag to the registry
      # by minimizing the shuffling of content into different layers
      - name: Check Base Image Availability
        id: check-baseline
        env:
          BASELINE_IMAGE: ${{ env.IMAGE_REGISTRY }}/${{ env.IMAGE_NAME }}:${{ env.DEFAULT_TAG }}
        run: |
          # TODO: check if this pulls stable-daily when we run stable-daily and stable for weeklies
          # checks if image is rechunked here:
          # https://github.com/coreos/rpm-ostree/blob/653e6ee0ba5953665b2bc2606bd6ab89c3871454/rust/src/compose.rs#L511-L519
          if skopeo inspect docker://"${BASELINE_IMAGE}" | jq -e '.LayersData[1:] | all(.Annotations?["ostree.components"]?)'; then
            echo "found ${BASELINE_IMAGE} in registry."
            echo "BASELINE_IMAGE=${BASELINE_IMAGE}" >> "$GITHUB_OUTPUT"
            echo "PULL_BASELINE_IMAGE=true" >> "$GITHUB_OUTPUT"
          else
            echo "no baseline image with tag $BASELINE_IMAGE found in registry"
            BASELINE_IMAGE="localhost/${{ env.IMAGE_NAME }}:${{ env.DEFAULT_TAG }}"
            echo "BASELINE_IMAGE=${BASELINE_IMAGE}" >> "$GITHUB_OUTPUT"
            echo "PULL_BASELINE_IMAGE=false" >> "$GITHUB_OUTPUT"
          fi

```

renner0e: image

PR #2007, opened 2026-04-05
feat(ci): avoid intermediary disk writes with --layers

renner0e:
This should speed up builds a bit as layers are only commited once to disk instead of for every instruction in the Containerfile.


renner0e:
[with new change](https://github.com/ublue-os/aurora/actions/runs/24009796128/job/70019420433?pr=2007)
[old](https://github.com/ublue-os/aurora/actions/runs/23992324814/job/69974088071)

aurora latest 9m 59s ->8m 15s
aurora-dx latest 14m 35s ->13m 25s

nvidia-dx stable 16m 14s ->14m 23s

we can merge this with 26.04 runners, we seem to be hitting some strange bug in podman

```
Writing manifest to image destination
--> 33176b8c401a
Successfully tagged localhost/aurora-dx:latest
Error: identifier is not an image
```

other images work fine???

inffy (approved):

PR #2060, opened 2026-04-16
feat: add a couple language packs for spectacle OCR

renner0e:
To avoid us adding *all* 100+ packages to the image (each one is a couple MiB) I only added the most popular languages according to Flathub. See: https://flathub.org/statistics

If one is missing then people can just request it, at least it's getting used then.

Supersedes: https://github.com/ublue-os/aurora/pull/1798


inffy (approved):
finnish is in there so I'll accept

PR #2062, opened 2026-04-16
chore(beta): update common image in beta

inffy:
nvidia service fix


inffy:
nvm, just realised the nvidia service is not on beta

PR #2166, opened 2026-05-07
chore: remove uupd override

Lumaeris:
We can do it with the config instead!

Required: https://github.com/get-aurora-dev/common/pull/175


renner0e:
This is fine by me, but I'm kinda hesitant at the moment to merge this, we will make a stable release soon and I don't think we should touch our updater right before that.

inffy:
We can postpone this to next week after stable f44 is done

inffy (approved):

PR #2181, opened 2026-05-10
feat: add chunkah to justfile

renner0e:
We can easily experiment with this now and implement the rest when it makes sense.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -274,7 +274,6 @@ build-pipeline image="aurora" tag="latest" flavor="main" kernel_pin="":
 
 # Rechunk Image
 [group('Image')]
-[private]
 rechunk $image="aurora" $tag="latest" $flavor="main" ghcr="0" pipeline="0" previous_build="0":
     #!/usr/bin/bash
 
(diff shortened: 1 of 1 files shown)


renner0e:
getting funny issues on this one....

image

I think this may be related to not having /ostree (I am also using unified storage)

Lumaeris:
Sharing my findings but using Bazzite KDE/GNOME as a base. Unified storage is not used, at least I don't know if it was enabled ever

It appears that it generally works after `rpm-ostree rebase` and `bootc switch` on my end, even package layering didn't break:

Screenshot From 2026-05-10 11-49-29

Screenshot From 2026-05-10 13-00-29

I did a prune of both /ostree/ and /sysroot/ (with a slash at the end) using Chunkah (it suggested to prune just /sysroot/), rpm-ostree seems to restore ostree content on live system no matter which command was used to rebase to the image with Chunkah (https://github.com/Lumaeris/laser/blob/da0ac77d651bca74946f637c4f229da71518dcaa/.github/workflows/build-system.yaml#L102-L115).

I want to see if clean install from ISO with Chunkah applied will break somehow but I doubt it will

Also noting that I saw that bootc RPM was updated recently on :testing tag, I used :stable here. Probably bootc now has a regression like above, not sure.

Lumaeris:
> I want to see if clean install from ISO with Chunkah applied will break somehow but I doubt it will

It *feels* like it deploys the system a bit slowly now but after it's done then it just works (TM)

Screenshot_silverblue_2026-05-10_14:51:50

Screenshot_silverblue_2026-05-10_15:05:21

Pruning /sysroot/ doesn't break anything, ostree backend restores the content if needed albeit slowly. Pruning of /ostree/ is probably not needed, it's just a symlink to /sysroot/ostree

Lumaeris:
... not sure what happened but after removing Waydroid as a test I now have to basically redownload the whole image.

Screenshot From 2026-05-10 18-05-53

inffy (approved):

PR #2184, opened 2026-05-10
chore: cleanup of packages.sh

renner0e:
see commits


renner0e:
tested with bootc installing the image, flathub is still there by default.

inffy (approved):

PR #2186, opened 2026-05-10
feat: one dnf transaction for kernel swap

renner0e:
This is faster due to less transactions


renner0e:
not worth it, needs some more changes

PR #2190, opened 2026-05-10
feat(ci): preparation for testing branches 

renner0e:
work being done here:
https://github.com/get-aurora-dev/aurora-test/issues/13


renner0e:
Just to clarify, my proposition here is to only build stable in PR runs and not publishing those images.

We need some way to find out if the next scheduled stable run is gonna work. This is ugly we need something better.

renner0e:
Currently aliasing stable-daily to point to stable, effectively lying about the daily part, is kinda the only option I see here as we can't really put those people on latest due to ZFS not being a thing there and we of course can't put them on testing.

renner0e:
Using coreos-testing on :testing is not something I'm 100% set on, I'm open to ideas, the choice depends on how much we care about not having ZFS on the testing images. We will drop it in Fall anyway.

inffy:
> Using coreos-testing on :testing is not something I'm 100% set on, I'm open to ideas, the choice depends on how much we care about not having ZFS on the testing images. We will drop it in Fall anyway.

was thinking that would be good to have the testing using the latest fedora kernel, to maybe track for possible regressions. But not sure if this would be needed and maybe it would be better to just track the coreos-testing as that would be the kernel that would go to stable.

we can set it to coreos-testing for now (if we want to support zfs - but don't think we have so many zfs users which propalby would run `testing`) and once zfs is gone we can change it if we want.

Also as discussed we might want to drop the `stable-daily` at some point. this would need a communication post - maybe with F45 - or just communicate it that it will be pointed to stable and will be dropped at point X.

renner0e:
> was thinking that would be good to have the testing using the latest fedora kernel, to maybe track for possible regressions. But not sure if this would be needed and maybe it would be better to just track the coreos-testing as that would be the kernel that would go to stable.

Yeah we can also just change it to whatever we need at the time, that would be fine with me, no ZFS on testing then.

renner0e:
> Also as discussed we might want to drop the stable-daily at some point. this would need a communication post - maybe with F45 - or just communicate it that it will be pointed to stable and will be dropped at point X.

I do not like the idea of dropping tags, we can easily setup a compatibility tag to :latest after we have dropped ZFS, it doesn't cost us much to keep this forever. If we want to we can rebase users to the proper tag when the time comes.

renner0e:
this got

PR #2191, opened 2026-05-10
test: remove libzfs

renner0e:
libzfs7 is now available, I think it's good enough to check for the kmod being there to ensure we actually have ZFS installed.


renner0e:
this only applies to stable images, which we are currently not building anyway

PR #2199, opened 2026-05-12
fix(ci): push production tags after signing

### LINKED ISSUE #2135: Update Error Started Today in Aurora
### Describe the bug

This error just started today on updates:

➜ ujust update
note: automatic updates (stage) are enabled
Pulling manifest: ostree-image-signed:docker://ghcr.io/ublue-os/aurora-dx:stable-daily
error: Preparing import: Fetching manifest: failed to invoke method OpenImage: A signature was required, but no signature exists

I expected 2 + 2 to equal 4, but instead 2 + 2 equaled 6!

### Output of `bootc status`

```shell

```

### Output of `groups`

```shell

```

_No response_

renner0e:
If the signing step fails, causes for this can be ghcr jank or a
sigstore outage and not testing our changes properly beforehand. Then we
have already pushed images to the registry with production tags that
users can update to.

The clients won't actually update to these images as we enforce
verification on the client side to pull images from ghcr.io/ublue-os/
only with a valid signature attached. Otherwise this leads to an "update
failed" on the users POV, which is very confusing.

To workaround this we push a tag to the registry first that we don't
care about and clients don't check against for updates. I chose the job
id because it always changes.

We used to determine the digest one additional skopeo inspect call, this
should eliminiate a little bit of jank on that side as well as we are
now reading it from a digestfile instead.

This also happens to get rid of the action we use for retries to
workaround ghcr jank, we were forced to do this in the past because the
old podman version we used at the time didn't have those flags.

fixes: https://github.com/ublue-os/aurora/issues/2135

Related: https://github.com/ublue-os/bazzite/pull/4902
Related: https://github.com/ublue-os/main/issues/643
Related: https://github.com/ublue-os/bazzite/issues/3044

Blocked by: https://github.com/ublue-os/aurora/pull/2208, although trivial to fix 

this is more conservative, an alternative approach would be to not use cosign and use podman push instead, which would get rid off some complexity at least.
- https://github.com/zirconium-dev/zirconium/pull/261

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -720,6 +720,9 @@ push-image $image="aurora" $tag="latest" $flavor="main" $ghcr="0" $image_registr
 
       elif [[ "${temp_push}" == "1" ]]; then
         ${PUSH_CMD} ${image_name}:${tag} ${image_registry}/${image_name}:${tag}-${temp_push_tag}
+        # We need to push twice to workaround https://github.com/containers/podman/issues/27796
+        # If we don't do this then the digest changes and we are only signing this specific tag
+        ${PUSH_CMD} ${image_name}:${tag} ${image_registry}/${image_name}:${tag}-${temp_push_tag}
       fi
 
     elif [[ "{{ ghcr }}" == "0" ]]; then
--- a/Justfile
+++ b/Justfile
@@ -690,11 +690,18 @@ setup-cache $image="aurora" $tag="latest" $flavor="main" $ghcr="0" $github_event
 
     echo "${CACHE_NAME}" "${ALLOW_CACHE_WRITE}"
 
+# Examples:
+# just push-image aurora latest main 1 ghcr.io/renner0e renner0e 1 "$(cat /path/to/my.key)
+
 # Push Image to Registry
 [group('Utility')]
-push-image $image="aurora" $tag="latest" $flavor="main" $ghcr="0" $image_registry="" $temp_push="0" $temp_push_tag="0":
+push-image $image="aurora" $tag="latest" $flavor="main" $ghcr="0" $image_registry="" $repo_owner="" $sign="0" $cosign_private_key="":
     #!/usr/bin/bash
     set -eoux pipefail
+
+    image_name=$({{ just }} image_name '{{ image }}' '{{ tag }}' '{{ flavor }}')
+
+    alias_tags=$({{ just }} generate-build-tags '{{ image }}' '{{ tag }}' '{{ flavor }}')
 
     PUSH_CMD_ARGS=()
     PUSH_CMD_ARGS+=("--digestfile=/tmp/digestfile")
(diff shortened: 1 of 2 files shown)


renner0e:
working run: https://github.com/renner0e/aurora/actions/runs/25874561798/job/76038247798

Justfile
```diff
@@ -690,6 +690,41 @@ setup-cache $image="aurora" $tag="latest" $flavor="main" $ghcr="0" $github_event
...
+    PUSH_CMD_ARGS+=("--retry-delay=30s")
+    PUSH_CMD_ARGS+=("--retry=5")
+
+    PUSH_CMD=""${PODMAN}" push "${PUSH_CMD_ARGS[@]}""
+
+    image_name=$({{ just }} image_name '{{ image }}' '{{ tag }}' '{{ flavor }}')
+
+    alias_tags=$({{ just }} generate-build-tags '{{ image }}' '{{ tag }}' '{{ flavor }}')
+
+    if [[ "{{ ghcr }}" == "1" && -n "${image_registry}" && "${temp_push}" == "0" ]]; then
```
ledif: Can we restructure this `if` to instead nest it like:

```bash
if [[ "{{ ghcr }}" == "1"  ]]; then
    if [[ ${temp_push}" == "0" ]]; then
    # ...
    else
    # ...
    fi
else
    # ...
fi
```
Right now, it's a little awkward repeating `"{{ ghcr }}" == "1"` and there is no matching branch when ` "{{ ghcr }}" == "1"` but ` -n "${image_registry}"` is false

m2Giles:
https://github.com/ublue-os/akmods/blob/20d23e9e3c413701e2b311044d55df0c9a38a640/Justfile#L414

You can sign at the same time as pushing with podman/skopeo.

renner0e:
> https://github.com/ublue-os/akmods/blob/20d23e9e3c413701e2b311044d55df0c9a38a640/Justfile#L414
> 
> You can sign at the same time as pushing with podman/skopeo.

@m2Giles 

I did implement that now, it works for me, I can rebase and pull the image and I could confirm previously that I can only pull signed images from my namespace

https://github.com/renner0e/aurora/actions/runs/25931347084/job/76225716893

https://github.com/renner0e/aurora/attestations/27657953

20+ minutes for pushing is not good tho, have to check out why that is exactly

You can verify that the policy you setup works by pulling

`ghcr.io/renner0e/aurora:chunkah`, which is not signed and should therefore not pull

/etc/containers/registries.d/renner0e.yaml
```yaml
docker:
  ghcr.io/renner0e:
    use-sigstore-attachments: true
```

/etc/containers/policy.json
```json
            "ghcr.io/renner0e": [
                {
                    "type": "sigstoreSigned",
                    "keyPaths": ["/etc/pki/containers/ghcr.io-renner0e.pub"],
                    "signedIdentity": {
                        "type": "matchRepository"
                    }
                }
            ],
```

/etc/pki/containers/ghcr.io-renner0e.pub
```
-----BEGIN PUBLIC KEY-----
MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEex4fLC8Eb0S10Rm1XshUnM9JS0Xj
IA+TkwSCa2H/qMM7BIIKDeiD9nYv03dWCK2aqRWHCtMZYR1G6jfG4//dMA==
-----END PUBLIC KEY-----
```

m2Giles:
That's very, very slow.

renner0e:
> That's very, very slow.

The more I spend time on this, the more I prefer using the temporary tag approach.

as of the most recent commit, using

https://github.com/renner0e/aurora/actions/runs/25941016510/job/76258296788

cosign verify --key renner.pub ghcr.io/renner0e/aurora:latest

```
Error: no matching signatures: signature not found in transparency log
signature not found in transparency log
signature not found in transparency log
signature not found in transparency log
signature not found in transparency log
signature not found in transparency log
signature not found in transparency log
signature not found in transparency log
signature not found in transparency log
signature not found in transparency log
signature not found in transparency log
signature not found in transparency log
error during command execution: no matching signatures: signature not found in transparency log
signature not found in transparency log
signature not found in transparency log
signature not found in transparency log
signature not found in transparency log
signature not found in transparency log
signature not found in transparency log
signature not found in transparency log
signature not found in transparency log
signature not found in transparency log
signature not found in transparency log
signature not found in transparency log
```

so I have to figure out how to get it to upload this to rekor

renner0e:
closing in favor of:

https://github.com/ublue-os/aurora/pull/2231

Issue #1833, opened 2026-03-03
increment tag suffix for multiple daily builds


renner0e:
This got exposed by the broken changelogs. We need 2 stable builds with the new attached SBOM location. We hoped to see the changelog generation working by rebuilding another stable image which happened on the same day resulting in the same tag `stable-20260303`

The actual underlying issue is that we don't have an index at the end which differentiates the images

So if it's a new date -> tag `20260303` and additionally `20260303.1`

If we rebuild another image on the same day and `20260303.1` already exists we increment the last number by one, resulting in `20260303.2` 

Already implemented in main:

https://github.com/ublue-os/main/blob/9dde1d5bbe3920031cf470e1c51331bd163d844f/Justfile#L232

PR #2208, opened 2026-05-14
feat(ci): stop building stable-daily images

renner0e:
These do not make super much sense, this is in preparation for the testing branches so we don't build too many images.

We will still build stable images in PRs to find out if things are broken on those images only.

To not break existing users, point this to `:stable` instead, effectively lying about the cadence now. We have to do it this way because stable-daily has ZFS and `:latest` doesn't to avoid breaking that use case. In the future when we have removed ZFS we can point this to `:latest`.

this makes this much easier to implement:
- https://github.com/ublue-os/aurora/issues/1833

- [ ] remove tag from rebase-helper (https://github.com/get-aurora-dev/common/pull/177)
- [ ] handle kcm_ublue (https://github.com/ledif/kcm_ublue/issues/22)
- [ ] reflect this change in the docs (https://github.com/ublue-os/aurora-docs/pull/373)
- [ ] make a systemd service to rebase people from :stable-daily to :stable (kinda overkill and optional, compatibility tag we will have to keep forever anyway )
- [ ] make a blog post about the change

anything else we should be doing?

### DIFF
--- a/.github/changelogs.py
+++ b/.github/changelogs.py
@@ -453,8 +453,6 @@ def generate_changelog(
         curr_pretty = re.sub(r"\.\d{1,2}$", "", curr)
         # Remove target- from curr
         curr_pretty = re.sub(rf"^[a-z]+-|^[0-9]+-", "", curr_pretty)
-        if target == "stable-daily":
-            curr_pretty = re.sub(rf"^[a-z]+-", "", curr_pretty)
         if not fedora_version + "." in curr_pretty:
             curr_pretty=fedora_version + "." + curr_pretty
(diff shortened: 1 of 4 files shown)


inffy:
Added a point that we make a short blog post when we do this. I don't think we need to advertise that before hand that much

renner0e:
I want to get this merged in the next few days, I don't think the kcm and systemd service should be hard blockers as we will not get rid of the compatibility tag ever.

inffy:
Is this good to go or are we still waiting on something?

inffy (approved):

Issue #2152, opened 2026-05-03
Virt-manager cannot create/start new VMs


inffy:
<img width="358" height="108" alt="image" src="https://github.com/user-attachments/assets/940aa1ee-8232-49d7-9092-98211581f000" />

Details say: "  File "/usr/share/virt-manager/virtManager/object/domain.py", line 1446, in startup
    self._backend.create()
    ~~~~~~~~~~~~~~~~~~~~^^
  File "/usr/lib64/python3.14/site-packages/libvirt.py", line 1395, in create
    raise libvirtError('virDomainCreate() failed')
libvirt.libvirtError: operation failed: swtpm died and reported: libvirt:  error : cannot execute binary /usr/bin/swtpm: Permission denied"

SELinux gives

```
Apr 29 12:40:00 hogwart setroubleshoot[14909]: SELinux is preventing rpc-virtqemud from entrypoint access on the file /usr/bin/swtpm. For complete SELinux messages run: seal>
Apr 29 12:40:00 hogwart setroubleshoot[14909]: SELinux is preventing rpc-virtqemud from entrypoint access on the file /usr/bin/swtpm.
                                               
                                               *****  Plugin restorecon (99.5 confidence) suggests   ************************
                                               
                                               If you want to fix the label. 
                                               /usr/bin/swtpm default label should be swtpm_exec_t.
                                               Then you can run restorecon. The access attempt may have been stopped due to insufficient permissions to access a parent direc>
                                               Do
                                               # /sbin/restorecon -v /usr/bin/swtpm
                                               
                                               *****  Plugin catchall (1.49 confidence) suggests   **************************
                                               
                                               If you believe that rpc-virtqemud should be allowed entrypoint access on the swtpm file by default.
                                               Then you should report this as a bug.
                                               You can generate a local policy module to allow this access.
                                               Do
                                               allow this access for now by executing:
                                               # ausearch -c 'rpc-virtqemud' --raw | audit2allow -M my-rpcvirtqemud
                                               # semodule -X 300 -i my-rpcvirtqemud.pp
```

_Originally posted by @manfredlotz in https://github.com/ublue-os/aurora/discussions/2127#discussioncomment-16755111_

We have the swtpm-workaround.service which for some reason doesn't seem to set the correct labels on boot.

AFter boot, restarting the service sets the label correct for /usr/bin/swtpm

Before:
```
ls -laZ /usr/bin/swtpm
[...]

PR #2220, opened 2026-05-16
fix(swtpm-workaround): order after rechunker-group-fix.service

elgorrion:
## Problem

On Aurora-DX, VMs started with a vTPM after boot fail with:

```
swtpm died and reported: libvirt: error : cannot execute binary /usr/bin/swtpm: Permission denied
```

Tracked as #2152 (Aurora) and ublue-os/bluefin#4538 (Bluefin, byte-identical `swtpm-workaround.service` from `get-aurora-dev/common`). Only Aurora hits the boot-time failure.

## Root cause

`rechunker-group-fix.service` (Aurora-only) ends `ExecStart` with:

```
ExecStart=systemd-tmpfiles --create --remove --boot --exclude-prefix=/dev
```

`systemd-tmpfiles --create` re-applies ownership, mode, and the path-based default SELinux context on every invocation. The `C`-rule in `/usr/lib/tmpfiles.d/swtpm-workaround.conf`:

```
C /usr/local/bin/overrides/swtpm - - - - /usr/bin/swtpm
```

`-` means default context. `matchpathcon /var/usrlocal/bin/overrides/swtpm` → `bin_t` (no `swtpm_exec_t` fcontext rule for `/var/usrlocal/...`). Every `systemd-tmpfiles --create` pass relabels the override back to `bin_t`; the bind mount propagates that to `/usr/bin/swtpm`.

Both units only depend on `local-fs.target`, so they start in parallel. When `rechunker-group-fix.service` finishes after `swtpm-workaround.service`, the relabel is clobbered.

Post-boot reproduction:

```
$ sudo systemctl restart swtpm-workaround.service
$ ls -lZ /usr/local/bin/overrides/swtpm
... system_u:object_r:swtpm_exec_t:s0 ... swtpm

$ sudo systemd-tmpfiles --create --remove --boot --exclude-prefix=/dev
$ ls -lZ /usr/local/bin/overrides/swtpm
... system_u:object_r:bin_t:s0 ... swtpm
```

## Fix

Drop-in ordering `swtpm-workaround.service` after `rechunker-group-fix.service`. Placed in this repo rather than `get-aurora-dev/common` because the race is Aurora-only — Bluefin and other downstreams that consume `common` don't have this conflict.

## Verification

`aurora-dx:latest-44.20260516.1`, before and after:

- Without drop-in: both units finish around the same time at boot; `/usr/bin/swtpm` ends up `bin_t`; vTPM VMs fail.
- With drop-in: `rechunker-group-fix.service` finishes first, `swtpm-workaround.service` runs after, `/usr/bin/swtpm` stays `swtpm_exec_t`; AVC gone.

`systemd-analyze verify swtpm-workaround.service` clean.

Refs ublue-os/aurora#2152, ublue-os/bluefin#4538.

### DIFF
--- a/system_files/shared/usr/lib/systemd/system/swtpm-workaround.service.d/10-order-after-rechunker-group-fix.conf
+++ b/system_files/shared/usr/lib/systemd/system/swtpm-workaround.service.d/10-order-after-rechunker-group-fix.conf
@@ -1,0 +1,2 @@
+[Unit]
+After=rechunker-group-fix.service


inffy:
Thanks on working on this. We will go with a little more simpler fix this time https://github.com/ublue-os/aurora/pull/2223

PR #2223, opened 2026-05-16
fix: selinux issues regarding swtpm

renner0e:
see commit


inffy (dismissed):

inffy:
tested locally and works, swtpm has the correct label after this

inffy (approved):

PR #2231, opened 2026-05-17
fix(ci): push production tags after signing

### LINKED ISSUE #2135: Update Error Started Today in Aurora

renner0e:
If the signing step fails, causes for this can be ghcr jank or a sigstore outage and not testing our changes properly beforehand. Then we have already pushed images to the registry with production tags that users can update to.

The clients won't actually update to these images as we enforce verification on the client side to pull images from ghcr.io/ublue-os/ only with a valid signature attached. Otherwise this leads to an "update failed" on the users POV, which is very confusing. This has been a longstanding issue across the org [1] [2] [3].

To workaround this, we push a tag to the registry first that we don't care about and clients don't check against for updates. I chose the job id because it always changes.

This also happens to get rid of the action we use for retries to workaround ghcr jank, we were forced to do this in the past because the old podman version we used at the time didn't have those features.

Using podman to push signatures and images at the same time is not something that is viable at the moment unfortunately but is definitely the approach we should be using once it works well.

This effort can be seen at [4]. Although we are likely hitting [5] here where all layers get invalidated because of this bug. We are effectively pushing full images/no shared layers every time we do podman push. Which causes the overall push step in CI to take upwards of 20 minutes.

fixes: https://github.com/ublue-os/aurora/issues/2135

[1]: https://github.com/ublue-os/bazzite/pull/4902
[2]: https://github.com/ublue-os/main/issues/643
[3]: https://github.com/ublue-os/bazzite/issues/3044
[4]: https://github.com/containers/podman/issues/27796
[5]: https://github.com/ublue-os/aurora/pull/2199

### DIFF
--- a/.github/workflows/reusable-build.yml
+++ b/.github/workflows/reusable-build.yml
@@ -89,7 +89,7 @@ jobs:
 
       # FIXME: Implement retries for stuff like this
       - name: Install Cosign
-        uses: sigstore/cosign-installer@7e8b541eb2e61bf99390e1afd4be13a184e9ebc5 # v4.1.2
+        uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6 # v4.1.2
         if: github.event_name != 'pull_request'
 
       - name: Check Just Syntax
--- a/.github/workflows/reusable-build.yml
+++ b/.github/workflows/reusable-build.yml
@@ -433,7 +433,7 @@ jobs:
           IMAGE_NAME: ${{ env.IMAGE_NAME }}
           SBOM_DIGEST: ${{ steps.upload-sbom.outputs.sbom_digest }}
         run: |
-          cosign sign -y --key env://COSIGN_PRIVATE_KEY ${IMAGE_REGISTRY}/${IMAGE_NAME}@${SBOM_DIGEST}
+          cosign sign -y --use-signing-config=false --key env://COSIGN_PRIVATE_KEY ${IMAGE_REGISTRY}/${IMAGE_NAME}@${SBOM_DIGEST}
 
       - name: Attestation
         if: github.event_name != 'pull_request'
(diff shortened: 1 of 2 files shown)


.github/workflows/reusable-build.yml
```diff
@@ -323,6 +328,44 @@ jobs:
...
+      # Workaround bugs caused by pushing images but not signing them because of CI flakes
+      # FIXME: once https://github.com/containers/podman/issues/27796 resolved
+      - name: Push Ephemeral Tag
+        if: github.event_name != 'pull_request'
+        id: pre-push
+        env:
+          MATRIX_BASE_NAME: "${{ matrix.base_name }}"
+          MATRIX_IMAGE_FLAVOR: "${{ matrix.image_flavor }}"
+          MATRIX_STREAM_NAME: "${{ matrix.stream_name }}"
+          TEMP_PUSH_TAG: "${{ github.run_id }}"
```
renner0e: I'm not 100% decided on this, we can also just set this to "tmp" or something, might make it easier to clean up later, although I have no idea how we would cleanup a specific tag, it's just a "cosmetic" issue.

m2Giles: maybe just prepend staging or something for the actual tag name

renner0e: yeah I like this more [9e3e9f0](https://github.com/ublue-os/aurora/pull/2231/commits/9e3e9f08b994e96af093dc7f0534aedb5759004e)

inffy (dismissed):

renner0e:
yeah this was implemented to need
https://github.com/ublue-os/aurora/pull/2208

renner0e:
I'll give this one last go on a fork and then this is ready

renner0e:
this is ready now

inffy (approved):

PR #2233, opened 2026-05-18
chore: move to stable Sunshine package

inffy:
Maybe we could also just move recommending the flatpak. It was also updated and looking at few comments, looks like it might actually just work.


renner0e (approved):
SGTM for the time being, migrating people to the flatpak once that is proven to work well should be the future tho

PR #2250, opened 2026-05-22
feat(ci): implement testing stream workflows

renner0e:


renner0e:
should go through merge queue, only :testing should be built because PR against main

renner0e:
disabled the status checks for latest/stable on main

inffy (approved):

PR #2259, opened 2026-05-25
feat(ci): use chunkah for rechunking, rootless CI

renner0e:
see commits


renner0e:
I have no idea why bootc install is making up new layers

image

renner0e:
split this up:

https://github.com/ublue-os/aurora/pull/2349

PR #2273, opened 2026-05-29
feat: remove patched fwupd

renner0e:
There have been quite a couple changes in the last couple years that were bootloader related in Fedora, namely switching to bootupd [1] with F41 which allows updating the bootloader. Maybe this happenend to resolve the issue because it got changed to fedora at some point?

Originally needed because fwupd looks for EFI/our_ID/**.

The only change we make to our package is hardcoding efi_os_dir=fedora with a build option.

New installs have the directory set to fedora.

Introduced in 78671f2.

~~I have no idea if we still need this, this is a scream test.~~

[1]: https://fedoraproject.org/wiki/Changes/FedoraSilverblueBootupd

### DIFF
--- a/build_files/base/01-packages.sh
+++ b/build_files/base/01-packages.sh
@@ -180,14 +180,6 @@ if [[ "${#EXCLUDED_PACKAGES[@]}" -gt 0 ]]; then
     fi
 fi
 
-# https://github.com/ublue-os/bazzite/issues/1400
-# TODO: test if we still need this when upgrading firmware with fwupd
-dnf -y copr enable ublue-os/staging
-dnf -y copr disable ublue-os/staging
-dnf -y swap \
-  --repo=copr:copr.fedorainfracloud.org:ublue-os:staging \
-  fwupd fwupd
-
 ## Pins and Overrides
 ## Use this section to pin packages in order to avoid regressions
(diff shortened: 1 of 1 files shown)


renner0e:
all my machines have the efi directory as fedora, some of those installations spanning back 3 years, some of them more recent. I think we should at least try removing this to confirm we actually still need this. Not sure exactly what thing changed it from aurora|bazzite to fedora at some point.

```
sudo tree /boot/efi/EFI

/boot/efi/EFI
├── BOOT
│   ├── BOOTIA32.EFI
│   ├── BOOTX64.EFI
│   ├── fbia32.efi
│   └── fbx64.efi
└── fedora
    ├── BOOTIA32.CSV
    ├── BOOTX64.CSV
    ├── grub.cfg
    ├── grubia32.efi
    ├── grubx64.efi
    ├── mmia32.efi
    ├── mmx64.efi
    ├── shim.efi
    ├── shimia32.efi
    └── shimx64.efi
```

inffy (approved):

renner0e:
https://github.com/fwupd/fwupd/blob/862012c6c5f850ddcebbc5cdf5c96005475b3861/plugins/uefi-capsule/fu-uefi-capsule-device.c#L721-L732

Fedora does not hardcode their ID so we should be good!

PR #2315, opened 2026-06-04
fix: workaround for selinux mprotect on composefs

inffy:
Relates: https://github.com/ublue-os/akmods/issues/537

### DIFF
--- a/build_files/base/01-packages.sh
+++ b/build_files/base/01-packages.sh
@@ -146,6 +146,8 @@ dnf config-manager addrepo --from-repofile=https://pkgs.tailscale.com/stable/fed
 dnf config-manager setopt tailscale-stable.enabled=0
 dnf -y install --enablerepo='tailscale-stable' tailscale
 
+# NOTE: Remove ublue-os-selinux-workarounds package when upstream issue is fixed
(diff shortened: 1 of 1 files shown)


renner0e (changes requested):
This seems to be not an issue exclusive to Nvidia
https://github.com/ublue-os/akmods/issues/537#issuecomment-4623932178

inffy:
> This seems to be not an issue exclusive to Nvidia [ublue-os/akmods#537 (comment)](https://github.com/ublue-os/akmods/issues/537#issuecomment-4623932178)

sigh, of course i missed that comment then..ffs

build_files/base/01-packages.sh
```diff
@@ -150,6 +150,7 @@ dnf -y install --enablerepo='tailscale-stable' tailscale
 copr_install_isolated "ublue-os/packages" \
     "kcm_ublue" \
     "krunner-bazaar" \
+    "ublue-os-selinux-workarounds" \
```
renner0e, line 155: I think we should leave a comment with the issue somewhere

renner0e (approved):

PR #2367, opened 2026-06-11
chore: bump ubuntu runner to 26.04 for image build

renner0e:


renner0e:
this is a nice improvement, let's bring it to stable as well

renner0e:
seems like there are still some network related problems with the fedora mirrors only on 26.04 runners, we do not get those elsewhere, holding off for now

renner0e:
https://github.com/actions/runner-images/issues/14230 is resolved so let's bring this in finally

inffy (approved):

PR #2443, opened 2026-06-23
refactor(ci): initial work for using flag/arg-based workflow

### LINKED ISSUE #2436: consider args/flag-based workflow for recipes
currently we have some big recipes that accept many arguments only in a very specific order, for example to build an nvidia image one would have to do `just build aurora latest nvidia-open`, notice how we only wanted to change one value (the flavor) from the default but had to specify the image and the tag in that exact order?

This would fix that and turn this command into `just build --flavor/-f nvidia-open` and the rest is default values. This would also align more with like every other cli program where one can essentially give the flags in any order. Would also make things more readable in CI. Now you need to look up what the exact position of the two 1s mean to guess what it does?

https://github.com/ublue-os/aurora/blob/670b33981f30dd41dcd8e990d0eadd97b7847da7/.github/workflows/reusable-build.yml#L366-L373

https://github.com/ublue-os/aurora/blob/670b33981f30dd41dcd8e990d0eadd97b7847da7/Justfile#L805

https://github.com/casey/just#recipe-flags-and-options

Here is a simple example for the image_name recipe:

```diff
--- i/Justfile
+++ w/Justfile
@@ -554,6 +554,11 @@ akmods_flavor $tag="latest":
 
 # Image Name
 [private]
+[
+  arg ("image", long="image", short="i"),
+  arg ("tag", long="tag", short="t"),
+  arg ("flavor", long="flavor", short="f")
+]
 image_name image="aurora" tag="latest" flavor="main":
     #!/usr/bin/bash
     set -eou pipefail
```

```patch
--- i/.github/workflows/reusable-build.yml
+++ w/.github/workflows/reusable-build.yml
@@ -134,7 +134,11 @@ jobs:
           MATRIX_STREAM_NAME: ${{ matrix.stream_name }}
           MATRIX_IMAGE_FLAVOR: ${{ matrix.image_flavor }}
         run: |
-          IMAGE_NAME="$(just image_name ${MATRIX_BASE_NAME} ${MATRIX_STREAM_NAME} ${MATRIX_IMAGE_FLAVOR})"
+          IMAGE_NAME="$(just image_name \
+                      --image ${MATRIX_BASE_NAME} \
+                      --tag ${MATRIX_STREAM_NAME} \
+                      --flavor ${MATRIX_IMAGE_FLAVOR})"
+
           echo "IMAGE_NAME=${IMAGE_NAME}" >> $GITHUB_ENV
 
       - name: DNF Package Cache Setup
```

Of course this would also need changes on the workflows like everywhere

For things like the `ghcr` bit we could do `--ghcr`
```just
  arg ("ghcr", long="ghcr", value="true")
```

Maybe we could also kill off/change the image name argument, it would always be set to aurora, and then we add `--dx` to the recipe instead of `just build aurora-dx stable nviida-open`. But this likely means even more work and should be a separate thing and not super important. If it's too much pain then we simply don't do that.
This would mean the syntax for aurora-dx-nvidia-open:stable would be (the flags can be in any order): `just build --dx --flavor nvidia-open --tag stable `

renner0e:
see commits, we need to make sure we are actually building the right images here, re-validating like everything 

Please hammer at this, especially edge cases. This won't see any big changes anymore, feel free to review!

needs docs changes: 
- https://github.com/ublue-os/aurora-docs/pull/387

resolves: https://github.com/ublue-os/aurora/issues/2436

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -764,13 +764,16 @@ setup-cache $image="aurora" $tag="latest" $flavor="main" $ghcr="false" $github_e
 
     echo "${CACHE_NAME}" "${ALLOW_CACHE_WRITE}"
 
-[group('Utility')]
+[arg("flavor", long="flavor", short="f")]
+[arg("image", long="image", short="i")]
+[arg("tag", long="tag", short="t")]
 [private]
 bootc $image="aurora" $tag="latest" $flavor="main" *ARGS:
     #!/usr/bin/env bash
     set -eoux pipefail
 
-    image_name=$({{ just }} image_name {{ image }} {{ tag }} {{ flavor }})
+    {{ just }} validate --image "${image}" --tag "${tag}" --flavor "${flavor}"
+    image_name=$({{ just }} image_name --image "${image}" --tag "${tag}" --flavor "${flavor}")
 
     BOOTC_INSTALL_OPTIONS=()
     BOOTC_INSTALL_OPTIONS+=("-v" "/var/lib/containers:/var/lib/containers" "-v" "/etc/containers:/etc/containers")
@@ -788,13 +791,18 @@ bootc $image="aurora" $tag="latest" $flavor="main" *ARGS:
         "${image_name}:${tag}" bootc {{ ARGS }}
 
 # Create bootable image
+[arg("backend", long="backend")]
+[arg("flavor", long="flavor", short="f")]
+[arg("ghcr", long="ghcr", value="true")]
+[arg("image", long="image", short="i")]
+[arg("tag", long="tag", short="t")]
 [group('Utility')]
-disk-image $image="aurora" $tag="latest" $flavor="main" ghcr="0" $backend="ostree":
+disk-image $image="aurora" $tag="latest" $flavor="main" ghcr="false" $backend="ostree":
     #!/usr/bin/env bash
     set -eoux pipefail
 
     # this is enough for base aurora to make it more likely to run in CI
-    if [[ "{{ ghcr }}" == "1" ]]; then
+    if [[ "${ghcr}" == "true" ]]; then
       # absurd size so it will always be enough for the image
       IMG_SIZE=35G
     else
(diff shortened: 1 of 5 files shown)


Justfile
```diff
@@ -84,19 +85,22 @@ fix:
...
     rm -f changelog.md
     rm -f output.env
     rm -rf sbom_out
     rm -f /tmp/aurora*-tags.json
 
 # Check if valid combo
-[group('Utility')]
+[arg("flavor", long="flavor", short="f")]
+[arg("image", long="image", short="i")]
+[arg("tag", long="tag", short="t")]
```
renner0e, line 98: if anybody knows a trick to set these globally on a recipe basis, so this is not duplicated in like *every* recipe then let me know

Justfile
```diff
@@ -1,3 +1,4 @@
+# Min just version 1.46.0 https://github.com/casey/just/issues/2290
```
renner0e, line 1: I tried the solution presented in: https://github.com/casey/just/issues/2290#issuecomment-3304994239, this does still work, however, in reality this is not actually useful as any just command will complain about the new unknown syntax instead of evaluating this command first and exiting more gracefully.

renner0e:
manually triggered workflows on fork:

testing: https://github.com/renner0e/aurora/actions/runs/28021713959
latest: https://github.com/renner0e/aurora/actions/runs/28021724613
stable: https://github.com/renner0e/aurora/actions/runs/28021743623

Justfile
```diff
@@ -120,26 +124,56 @@ validate $image $tag $flavor:
...
+    ARCH=$(arch)
+
+    case "${tag}" in
+            stable)
+                if [[ "${ARCH}" == "x86_64" ]]; then
+                    # <Here is a link why we have it pinned>
+                    kernel_pin=""
+                elif [[ "${ARCH}" == "aarch64" ]]; then
+                    kernel_pin=""
+                fi
```
renner0e, line 159: not actually tested the ARM portions here

inffy:
Not sure what else to test but atleast local building works

```
just build -i aurora-dx -t testing
just build -i aurora-dx -t testing -f nvidia-open
```

renner0e:
forgot to mark this as ready for review

inffy (approved):

PR #2466, opened 2026-06-28
feat(ci): package cache with OCI artifacts

### LINKED ISSUE #2351: find a way to generate new package cache
As we are no longer running a scheduled workflow from the main branch due to implementing testing streams and a bunch of stuff changing because of that, we will not get a new package cache.

Maybe we should have a small workflow that runs on a schedule, that doesn't publish any images, it's just there to generate cache every week.

This will not work due to it only ever actually running on the merge queue branch and not main, where all branches get it's cache from:

make the "decide if we write cache"-logic more tight and allow it to run on the merge queue, [which is currently not possible](https://github.com/actions/cache/issues/1537), but only if it's for example a Tuesday between 5-7AM UTC+2 (or whenever renovate usually picks up the base image update)

renner0e:
We are hitting a couple limitations in regards to actions/cache [1] with our new testing branch workflow, as we previously relied on scheduled builds from the main branch, which we no longer do with this model.

Now we make new cache on merge_group on the first build on Sundays, I chose Sundays just because. Which means we run only a single build with no cache per week and use that same build to generate a new fresh cache.

Else we would end up with 3 versions of vscode in our cache. There is no point uploading cache on *every* PR.

For simplicity sake this has no handling/differentiation for our stable and main branch. When we bump the fedora version of our main/testing images then the cache for stable will just get stale.

We are mixing root/rootless usage of oras/cosign, I made it use root or else we will get premission issues.

[1]: https://github.com/actions/cache/issues/1537

fixes: https://github.com/ublue-os/aurora/issues/2351

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -743,15 +743,22 @@ gen-sbom $image="aurora" $tag="latest" $flavor="main" $syft_cmd="syft":
 
     rm -rf "${ROOTFS}"
 
+# We are not using https://github.com/actions/cache because of:
+# https://github.com/ublue-os/aurora/issues/2351
+# https://github.com/actions/cache/issues/1537
+
 # DNF CI package cache
 [arg("flavor", long="flavor", short="f")]
 [arg("ghcr", long="ghcr", value="true")]
 [arg("github_event", long="github-event")]
 [arg("image", long="image", short="i")]
+[arg("pull", long="pull", value="true")]
+[arg("push", long="push", value="true")]
+[arg("registry", long="registry")]
 [arg("tag", long="tag", short="t")]
 [group('Utility')]
 [private]
-setup-cache $image="aurora" $tag="latest" $flavor="main" $ghcr="false" $github_event="":
+setup-cache $image="aurora" $tag="latest" $flavor="main" $ghcr="false" $pull="false" $push="false" $registry="" $github_event="":
     #!/usr/bin/env bash
     set -eou pipefail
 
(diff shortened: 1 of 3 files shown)


Justfile
```diff
@@ -743,35 +743,62 @@ gen-sbom $image="aurora" $tag="latest" $flavor="main" $syft_cmd="syft":
...
+
+    POINT=$({{ just }} generate-point --image "${image}" --tag "${tag}" --flavor "${flavor}")
+    CURRENT_DAY="$(LC_TIME=C date +%A)"
+
+    if [[ "${CURRENT_DAY}" == "Sunday" && "$POINT" == "1" && "${ghcr}" == "true" ]]; then
+      echo "Not pulling build cache. Uploading fresh cache later."
+    elif [[ "${pull}" == "true" ]]; then
+      # We want to avoid using --allow-path-traversal
+      pushd "${BUILDAH_CACHE_DIR}"
+      oras pull "${CACHE_IMAGE}" || exit 0
```
renner0e, line 789: should we sign/verify the cache? I **think** (please someone correct me) DNF only allows installation if the package is signed with a trusted key, which I think most are except for things we install "offline" from ublue-os/akmods, at least according to `rpm -qi` but it doesn't throw any warning or the like

renner0e:
works in my fork: https://github.com/renner0e/aurora/actions/runs/28334938260/job/83939369446

should better be merged next saturday so can see it working when it is supposed to.

inffy (dismissed):

inffy (approved):

PR #2474, opened 2026-06-30
chore: promote changes to stable

renner0e:
obsoletes: https://github.com/ublue-os/aurora/pull/2357

this PR is needed because of merge conflicts, we should look into making the pull app create a separate branch and this PR merges into stable, as we can't/shouldn't push directly to the stable/main branches and only make any changes via PRs


renner0e:
changed a setting, merges (and squashes) are allowed for the stable branch now

inffy (approved):

PR #2496, opened 2026-07-03
feat(ci): rootless CI

renner0e:
26.04 runners use sudo-rs, which does not support this flag. It doesn't
seem to have broken too much, Only thing I could see is the github token
not being passed to the build.

```
sudo: preserving the entire environment is not supported, '-E' is ignored
```

So this is a good time to move over to a rootless CI I guess.


renner0e:
disk image is failing, needs https://github.com/ublue-os/aurora/pull/2497 first

inffy (approved):

Issue #2492, opened 2026-07-03
Cannot install `firewall-config`


JeanLuX:
### Describe the bug

Hello,

I’m unable to install the [FirewallD graphical interface](https://firewalld.org/documentation/utilities/firewall-config.html) (`firewall-config`) on the current Aurora image.

`rpm-ostree` reports that the package is already provided by the base image:

`firewall-config (already provided by firewall-config-2.4.0-2.fc44.noarch)`

However, this does not seem to be the case in practice:
- the `firewall-config` binary is not present,
- the desktop entry is also missing,
- and the application cannot be launched after reboot.

Running `rpm-ostree install firewall-config` completes, but even after rebooting, `firewall-config` is still not installed or available in the system. Even though it can now be uninstalled using `rpm-ostree uninstall firewall-config` (which is essentially useless)

## Steps to reproduce

1. Boot the current Aurora image: `latest-44.20260703.2`
2. Run:
   ```bash
   rpm-ostree install firewall-config
   ```
3. Reboot the system.
4. Try to launch:
   ```bash
   firewall-config
   ```

## Actual behavior

`rpm-ostree` reports that `firewall-config` is already provided, but the binary is missing and the package is not usable after installation and reboot.

<img width="2273" height="1385" alt="Image" src="https://github.com/user-attachments/assets/8b6a9726-0b39-41d5-8959-6c5916510f79" />

### What did you expect to happen?

`firewall-config` should either:
- be actually present in the base image, or
- be correctly layered and available after `rpm-ostree install firewall-config`.

### Output of `rpm-ostree status`

```shell
State: idle
AutomaticUpdates: stage; rpm-ostreed-automatic.timer: inactive
Deployments:
● ostree-image-signed:docker://ghcr.io/ublue-os/aurora-dx:latest
Digest: sha256:e9be27c80ea28961b94381ac820c7466a6c0a6932972ed37a1d900482281cf39
Version: latest-44.20260703.2 (2026-07-03T08:24:06Z)
LocalPackages: openconnect-globalprotect-1.4.9-1.mga9.x86_64
InitramfsEtc: /etc/crypttab

ostree-image-signed:docker://ghcr.io/ublue-os/aurora-dx:latest
Digest: sha256:e9be27c80ea28961b94381ac820c7466a6c0a6932972ed37a1d900482281cf39
Version: latest-44.20260703.2 (2026-07-03T08:24:06Z)
[...]

Issue #2209, opened 2026-05-14
Missing firefox in stable-20260511


tuxPT:
### Describe the bug

When upgrading to latest stable-dx, firefox becomes an inactive request. However it doesn't exist in /usr/bin/firefox. Uninstalling it does not solve, nor either installing it(it just becomes inactive request again).

Only way to make use of it is to install through:
1. rpm-ostree usroverlay
2. dnf5 install firefox

From what I had read in packages.sh in the build-files directory, you exclude firefox from installation because it "supposedly" is already in the shared image. 

### What did you expect to happen?

firefox installed, /usr/bin/firefox existing.

renner0e:
Use the flatpak of firefox. This is likely a gotcha of the way rpm-ostree handles things.

> From what I had read in packages.sh in the build-files directory, you exclude firefox from installation because it "supposedly" is already in the shared image.

Where do you see that? We remove it from our base image because we want to use the official build from mozilla that is published to flathub..

tuxPT:
I don't want to use flatpak version. Browsers have their own sandboxes. Don't know what you had done, but the system thinks firefox is in the image when it isn't and does not allow me to install it either. Only affects this specific release(the fedora-44 one).

tuxPT:
https://bugzilla.mozilla.org/show_bug.cgi?id=1756236

https://seirdy.one/notes/2022/06/12/flatpak-and-web-browsers/

renner0e:
We have switched over to the upstream rechunker included in rpm-ostree with this release, instead of our own custom unmaintained one.

The way they handle things with the rpmdb seems to be different, we will switch over to a different newer rechunker in the next months, which I am using right now and I can layer firefox again. Although this is not a use case we will ever support.

>I don't want to use flatpak version. Browsers have their own sandboxes

We will never recommend layering browsers.

It is true that flatpak browsers generally have a weaker sandbox, although it is not a black and white issue as other solutions like distrobox and layering browsers on top of the image means extra maintenance work for you. You have to make sure that you setup auto-updates in case of distrobox for example! Where as the flatpak has a very low chance that auto updates break and you are not accidentally running an outdated version with known vulnerabilities for a prolonged amount of time.

Bundling browsers with the image is not a perfect solution either as in the case of rolling back your image, let's say an issue that breaks some of your hardware functionality on newer kernels, as you downgrade the entire image and not just individual components. This will also mean running an old browser version that came with that image.

[...]

PR #2499, opened 2026-07-03
fix: relink rpm-ostree-base-db to system rpmdb

renner0e:
taken from blue-build[1], this also happens to make updates smaller by
100MB. Might fix the following issues like [2] and [3]. I could not
reproduce these issues myself and just using chunkah made it possible
for me to overlay firefox with rpm-ostree (without changes this PR makes)

[1]: https://github.com/blue-build/cli/commit/b6f36bd9a2efff30adb06cf9dba43a24d6f47891
[2]: https://github.com/ublue-os/aurora/issues/2209
[3]: https://github.com/ublue-os/aurora/issues/2492


renner0e:
do not put this on the stable branch, not before we have confirmation that it doesn't break anything (layers disappear, updates break in any way)

inffy (approved):

PR #2503, opened 2026-07-04
fix(ci): fix the backport action to use the base ref

inffy:
Looks like by default the backport action tries to checkout the fork PR that was merged and that triggers the checkout actions security protections. Not sure why are we seeing this just only now.

This should make it use the base branch explicitly (main) to checkout as the PR is already merged into it.


renner0e (approved):
Awesome, let's try it!

Issue #2337, opened 2026-06-07
CI flakes tracker


renner0e:
base image verification fails
```
  Error: Get "https://quay.io/v2/": dial tcp 3.92.171.144:443: i/o timeout
  error during command execution: Get "https://quay.io/v2/": dial tcp 3.92.171.144:443: i/o timeout
  NOTICE: Verification failed. Please ensure your public key is correct.
  error: recipe `verify-container` failed with exit code 1
```

[base image pull fails](https://github.com/ublue-os/aurora/actions/runs/27457552087/job/81164795487)
```
fetching blob: received unexpected HTTP status: 502 Bad Gateway
```

creating attestation:
```
Error: InternalError: error creating signing certificate - (500) Error entering certificate in CTL
```

copr

```
+ dnf5 -y install --enablerepo=copr:copr.fedorainfracloud.org:ublue-os:packages kcm_ublue krunner-bazaar ublue-os-selinux-workarounds oversteer-udev uupd
  Updating and loading repositories:
   Copr repo for packages owned by ublue- 100% |  65.0   B/s |   2.7 KiB |  00m43s
  >>> Curl error (28): Timeout was reached for https://download.copr.fedorainfracloud.org/results/ublue-os/packages/fedora-44-x86_64/repodata/repomd.xml [Operation too slow. Less than 1000 bytes/sec transferred the last 30 seconds] - https://download.copr.fedorainfracloud.org/results/ublue-os/packages/fedora-44-x86_64/repodata/repomd.xml
  >>> Status code: 504 for https://download.copr.fedorainfracloud.org/results/ublue-os/packages/fedora-44-x86_64/repodata/repomd.xml (IP: 108.138.85.27) - https://download.copr.fedorainfracloud.org/results/ublue-os/packages/fedora-44-x86_64/repodata/repomd.xml
  >>> Status code: 504 for https://download.copr.fedorainfracloud.org/results/ublue-os/packages/fedora-44-x86_64/repodata/repomd.xml (IP: 108.138.85.27) - https://download.copr.fedorainfracloud.org/results/ublue-os/packages/fedora-44-x86_64/repodata/repomd.xml
  >>> Status code: 504 for https://download.copr.fedorainfracloud.org/results/ublue-os/packages/fedora-44-x86_64/repodata/repomd.xml (IP: 108.138.85.27) - https://download.copr.fedorainfracloud.org/results/ublue-os/packages/fedora-44-x86_64/repodata/repomd.xml
[...]

PR #2504, opened 2026-07-04
feat(ci): retries for pulled images

### LINKED ISSUE #2381: put image pulls in separate just recipe
This is so we can utilize retry functionality of podman when called with podman pull instead of implicitly being pulled by `FROM`. Should reduce flakiness at least a little bit.

renner0e:
The --retry flags for podman/buildah pull/build do not seem to do
anything on transient network issues. Notably our base image pull from
quay is failing quite often, this should make this less likely to
happen in the future.

fixes: https://github.com/ublue-os/aurora/issues/2381
xref: https://github.com/ublue-os/aurora/issues/2337

### DIFF
--- a/.github/workflows/reusable-build.yml
+++ b/.github/workflows/reusable-build.yml
@@ -163,7 +163,8 @@ jobs:
             --image "${MATRIX_BASE_NAME}" \
             --tag "${MATRIX_STREAM_NAME}" \
             --flavor "${MATRIX_IMAGE_FLAVOR}" \
-            --ghcr
+            --ghcr \
+            --retry-pull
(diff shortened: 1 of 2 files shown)


renner0e:
out of 108 images, none of them failed because of the quay pull, so i think this generally does what it is supposed to do
https://github.com/ublue-os/aurora/actions/runs/28703424141/job/85125133170?pr=2504

We will see if this actually proves to be effective

renner0e:
let's get this in already, jesus chris i saw like more than a handful of failures just today

inffy (approved):

Issue #2528, opened 2026-07-08
harden image build, code scanning, tighten workflow permissions tracker issue


renner0e:
- [ ] implement linters (mainly bash), figure out what to do about codacy (doesn't work on non-default branch)
- [ ] vulnerability scanning with trivy, evaluate if this actually makes sense for us
- [ ] [generally following best practices](https://scorecard.dev/viewer/?uri=github.com/ublue-os/aurora), 
- [ ] implement https://docs.zizmor.sh/
- [ ] investigate the scorecard regression check again
- [ ] monitor outgoing network requests
- [ ] question and document why we have certain github worklfow permissions

[Secureblue](https://github.com/secureblue/secureblue/tree/live/.github/workflows) generally does most of those things, we should take a look at that and see what we can improve/adapt to our usecase

ties a little bit into https://github.com/ublue-os/aurora/issues/2419

I think we should have a separate justfile that is only for linting and scanning

PR #2532, opened 2026-07-08
chore(ci): use built-in release function of gh

renner0e:
We should not rely on external actions if this functionality exists in official tools.

We are only running this step when it's a stable image build so ommiting the make_latest thing is fine to my understanding.

xref: https://github.com/ublue-os/aurora/issues/2528


renner0e:
a release I created locally with gh

https://github.com/renner0e/aurora/releases/tag/testing-20260708.3

inffy (approved):

Issue #2350, opened 2026-06-08
chunkah tracker issue


renner0e:
- https://github.com/ublue-os/aurora/pull/2349

- [ ] [implement passing previous manifest](https://github.com/coreos/chunkah/issues/39)
- [x] investigate `--oci` option, could be worth it would allow us to skip the export-oci step on PRs, maybe it's also slightly faster than `podman save | podman load`?

PR #2545, opened 2026-07-12
feat(chunkah): use oci-dir instead of tar-roundtrip

renner0e:
/tmp is a tmpfs backed by RAM (8G) on github so this will also be faster than the current implementation because of that.

See: https://github.com/coreos/chunkah#output-options

xref: https://github.com/ublue-os/aurora/issues/2350


renner0e:
just a rough comparison on how much this makes a difference:

aurora-dx
https://github.com/ublue-os/aurora/actions/runs/29163088990/job/86571412386
before: 11m 53s
after: 4m 45s

inffy (approved):

PR #2549, opened 2026-07-12
feat(ci): setup runner with containers policy

renner0e:
Currently for ublue-os/akmods and kinoite (not pinned by digest), we
verify them and pull them right after. We can't trust the authenticity
of those images, as the tag in the meantime could point to a different,
potentially compromised digest.

To avoid this, we set up our runner so that it will refuse to pull
anything from Universal Blue or quay.io/fedora-ostree-desktops without
first getting verified with the corresponding public key in our repo.

With this, we can completely get rid of our verify-container recipe and
benefit from the retry logic that is implemented in the podman pulls we
already do beforehand.

This would mean that local builds would no longer verify the pulled
images, as the policy is very likely not configured, but I don't think
that is super important.

xref: https://github.com/ublue-os/aurora/issues/2337

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -144,7 +144,6 @@ build $image="aurora" $tag="latest" $flavor="main" $rechunk="false" $ghcr="false
 
     BASE_IMAGE_REF="${base_image_org}/${base_image_name}:${fedora_version}"
     ALL_IMAGES=()
-    {{ just }} verify-container quay.io-fedora-ostree-desktops.pub "${BASE_IMAGE_REF}"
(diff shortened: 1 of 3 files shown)


Justfile
renner0e, line 211: oidc/keyless verification is currently not implemented in containers-libs, so we are stuck with cosign there

https://github.com/podman-container-tools/container-libs/issues/388
https://github.com/podman-container-tools/container-libs/issues/34

renner0e:
you can see this working here: https://github.com/renner0e/aurora/actions/runs/29208695754/job/86692575116

inffy (approved):

PR #2605, opened 2026-07-22
chore(ci): stop hardcoding stable-branch

renner0e:
Instead we can define an environment variable for github actions.

For things where we can't, globs/regex is probably good enough.

When we branch/stabilize "stable-f45" then we will delete the old one anyway, so we shouldn't be in a situation where we'll get spammed with like 3 PRs.

The only remaining things where it's hardcoded is the pull config and docs.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -4,7 +4,7 @@
     "config:best-practices"
   ],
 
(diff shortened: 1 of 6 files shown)


renner0e:
let's see if things blow up

inffy (approved):

Issue #1765, opened 2026-02-15
We will remove ZFS support starting with Fedora 45 based images in Fall 2026.


renner0e:
This is kinda awkward as the Fedora Kernels are rather aggressive (even with a 2 week coreos delay) which does not play well with out of tree kernel modules like ZFS. Which means we either have to use older kernels for a prolonged time (where zfs still works/has support) or don't publish images stable images for that period.

We don't even mention ZFS anywhere in our docs and website so most people don't even know that we have it, we also don't expose it in the installer and it is also not super useful in a desktop usecase.

This absolutely sucks that we have to remove it to make aurora more sustainable long-term, especially a filesystem that breaks everything if support for it is removed.

castrojo:
Stuck with the same problem, I'd like to announce deprecation in Bluefin soonish with removal in the Fall. Though I expect the ZFS support in LTS will remain fine for its entire lifecycle.

allanjude:
It looks like 6.18 was supported by OpenZFS on Dec 10th, and 6.19 on Feb 11th.
Which version of ZFS are you trying to aim to use?

Danathar:
> It looks like 6.18 was supported by OpenZFS on Dec 10th, and 6.19 on Feb 11th. Which version of ZFS are you trying to aim to use?

Thanks Allan! I knew listening to all those episodes of 2.5 admins might come in handy!

allanjude:
> We are using the 2.3 branch at the moment and have been waiting for 2.3.6 release so we can build the kmod.
> 
> https://github.com/ublue-os/akmods/tree/main/build_files/zfs Is where our akmods happen and it looks for github releases

2.3.6 was cut today with support for 6.19, and the packages are building now.

m2Giles:
> Also Bluefin LTS (and HWE) exists - I don't see a reason for us to stop supporting it there.

HWE depending on how you implemented might be affected. But Using the 6.12 Centos Kernel or the 6.12 LongTerm Fedora Kernel would not have the current issues.

itoffshore:
I've just started looking at Atomic Desktop (after Arch installed Wayland over X11 breaking everything) 
 
- I think if the **Cachyos LTS kernel** can be included as an option - `zfs` should always work (as they build `zfs` against all of their kernels)
- I came across today [Origami Linux](https://origami.wf/#features) - atomic desktop + cachyos kernel (not sure if `zfs` is possible there yet) - so potentially possible in Aurora too ?

[...]

PR #2613, opened 2026-07-24
feat: add ZFS deprecation notifier

### LINKED ISSUE #2210: implement notification service to warn ZFS users of deprecation
This should be done in a graphical way that is impossible to avoid, keep it simple, something with `kdialog` which will write a stamp file so it won't trigger again.

renner0e:
This will spawn a notification every week that has to be dismissed by the user if ZFS is being used.

This can be permanently disabled with:

```
sudo systemctl disable aurora-zfs-deprecation-notifier.timer
```

xref: https://github.com/ublue-os/aurora/issues/1765
resolves: https://github.com/ublue-os/aurora/issues/2210

### DIFF
--- a/system_files/shared/usr/bin/aurora-zfs-deprecation-notifier
+++ b/system_files/shared/usr/bin/aurora-zfs-deprecation-notifier
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+NOTIFICATION="/usr/bin/notify-send --app-name=Aurora --icon distributor-logo --urgency critical 'ZFS is getting removed in Fall 2026' 'You are using ZFS so please take a look at https://github.com/ublue-os/aurora/issues/1765 for more infos. There will not be another reminder.'"
+
+USERS="$(loginctl list-users --no-legend | awk '$1 != 0 {print $2}')"
+
+for username in ${USERS}; do
+  machinectl shell "${username}@.host" /usr/bin/bash -c "${NOTIFICATION}"
+done
+
+touch /etc/.aurora-zfs-deprecation-notifer.stamp
+
--- a/system_files/shared/usr/bin/aurora-zfs-deprecation-notifier
+++ b/system_files/shared/usr/bin/aurora-zfs-deprecation-notifier
@@ -2,7 +2,7 @@
 
 set -euo pipefail
 
-NOTIFICATION="/usr/bin/notify-send --app-name=Aurora --icon distributor-logo --urgency critical 'ZFS is getting removed in Fall 2026' 'You are using ZFS so please take a look at https://github.com/ublue-os/aurora/issues/1765 for more infos. There will not be another reminder.'"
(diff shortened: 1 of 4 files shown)


renner0e (commented):
to clarify how I want this to work:

- no ZFS (akmods-main) -> don't run anything
- no disk with ZFS -> also don't run anything
- (ideally) only create a stamp file when the notification appeared

People who use Aurora headless (for whatever reason :/) will probably not see anything here. The way it's currently implemented the timer should start after PLM, a desktop will probably be running then?

Image

renner0e:
this is fine now, it shouldn't be crazy annoying or notify people who don't care about ZFS, any more suggestions on wording?

EDIT: actually, how about we make this more annoying, make the timer run every every week/2 weeks (and don't ever write a stampfile)

system_files/shared/usr/bin/aurora-zfs-deprecation-notifier
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+NOTIFICATION="/usr/bin/notify-send --app-name=Aurora --icon distributor-logo --urgency critical 'ZFS is getting removed in Fall 2026' 'You are using ZFS so please take a look at https://github.com/ublue-os/aurora/issues/1765 for more infos. There will not be another reminder.'"
```
ledif:
```suggestion
NOTIFICATION="/usr/bin/notify-send --app-name=Aurora --icon distributor-logo --urgency critical 'ZFS support is ending Fall 2026' 'An attached drive contains a ZFS pool. Please see https://github.com/ublue-os/aurora/issues/1765 for more info.'"
```

ledif (commented):
Your previous comments suggest that your branch previously had a mechanism for the user to be reminded again. Can we have that option? Or do we think a one-time notification is sufficient?

It might be useful for a user that has a ZFS drive that they use regularly use to be able to communicate "not now, but please remind me again."

system_files/shared/usr/bin/aurora-zfs-deprecation-notifier
renner0e: I like this more.

Maybe we should say October (even if it ends up being November in reality) instead of being vague with fall?

renner0e:
> Your previous comments suggest that your branch previously had a mechanism for the user to be reminded again. Can we have that option?

Nope I have not had that mechanism previously, but we could easily do that if we just don't write the stampfile.

This approach would probably be the most annoying (reach more users!) and would be easy to do.

And if the user knows about it then they can just disable the timer to get rid of the weekly annoyance I guess.

system_files/shared/usr/bin/aurora-zfs-deprecation-notifier
renner0e: 79d0e8fefda1d22fc39b2049b97e21760aea88fd

ledif (approved):
> And if the user knows about it then they can just disable the timer to get rid of the weekly annoyance I guess.

Ah, good point. The fact that it's just a systemd timer that can be disabled is a pretty good design for this IMO.

LGTM.

PR #2619, opened 2026-07-25
feat: delete /run leftovers from the image

renner0e:
see commits


ledif (approved):
I didn't realize the contents of `/run` from the build were in the image!

renner0e:
> I didn't realize the contents of `/run` from the build were in the image!

*All* the files/folders were pretty much empty or very small, we are shipping the contents of them but bootc doesn't "deploy" them :laughing:

PR #2701, opened 2026-08-09
refactor: repurpose build.sh as build preparation

renner0e:
see commits


renner0e:
[diff-aurora.txt](https://github.com/user-attachments/files/30876520/diff-aurora.txt)

so looking at a diff here, seems to be mostly expected, all the no such file.... seem to be coming from now broken symlinks, which comes from the way I'm comparing them, those file actually do exist on both images.

The actual changes that matter that are not caused by build/testing churn:

```
Only in sbom_out/aurora_production/rootfs/etc/uupd: config.json.rpmnew
Only in sbom_out/aurora_new/rootfs/usr/share/plasma/plasma-welcome: intro-customization.desktop
Files sbom_out/aurora_production/rootfs/usr/share/qt5/qtlogging.ini and sbom_out/aurora_new/rootfs/usr/share/qt5/qtlogging.ini differ
```

qt5 actually has our config from common now, I guess we never noticed.
We don't have that nasty uupd rpmnrew file anymore which comes from -> copy config -> install package with said config, uupd is packaged properly so our configuration isn't being overwritten by the installation of it, this never caused issues in practice but still nice to see that fixed.

renner0e:
[diff-aurora-dx.txt](https://github.com/user-attachments/files/30876635/diff-aurora-dx.txt)

nothing of substance here, the differences of various libraries being different is because those packages got upgraded in between the time our image got built the last time compared to me building them.

Didn't test nvidia, but pretty certain that there should be nothing specific here to blow it up

renner0e:
by extension fixes: https://github.com/ublue-os/aurora/issues/2624

inffy (approved):

PR #2703, opened 2026-08-09
feat: set default applications via xdg-mime

### LINKED ISSUE #2658: "Email client" set in "Default applications" is a browser
### Describe the bug

Just started using Aurora, and noticed that the default application for an email client is a browser:

Image

I expected the email client to be set as Thunderbird by default, particularly so as it comes preinstalled in Aurora.

I am not entirely sure this is a bug, as there may be an actual reason for this. But in my mind, Thunderbird would make more sense.

### Output of `rpm-ostree status`

```shell
State: idle
AutomaticUpdates: stage; rpm-ostreed-automatic.timer: inactive
Deployments:
● ostree-image-signed:docker://ghcr.io/ublue-os/aurora-nvidia-open:stable
                   Digest: sha256:254cb08f9394450121b5ec870c51f3df23b5f7c76367b2021174c4c9874dfcb8
                  Version: 44.20260728.1 (2026-07-28T01:41:51Z)

  ostree-image-signed:docker://ghcr.io/ublue-os/aurora-nvidia-open:stable
                   Digest: sha256:ec5841d8c6d49ffdaf3415f485eb9cee4697cda88d0539704cfe4c794d3a88b1
                  Version: 44.20260714.1 (2026-07-14T01:40:34Z)
```

### Output of `groups`

```shell

```

_No response_

renner0e:
We need to do it this way because editing with sed/echo is nasty because we may end up with multiple keys of the same category/filetype.

No idea when this broke or this ever worked in the case of thunderbird, but it should now.

fixes: https://github.com/ublue-os/aurora/issues/2658

### DIFF
--- a/build_scripts/base/16-override-install.sh
+++ b/build_scripts/base/16-override-install.sh
@@ -22,9 +22,10 @@ rm -rf /usr/share/doc/HTML
 
 rm -f /usr/lib64/qt6/plugins/kf6/krunner/krunner_appstream.so
 
-# Use Bazaar for Flatpak refs
-# https://github.com/ublue-os/bazzite/pull/3620
-echo "application/vnd.flatpak.ref=io.github.kolunmi.Bazaar.desktop" >> /usr/share/applications/mimeapps.list
+# xdg-mime doesn't natively support using /usr/share/applications/mimeapps.list and would instead change /root/.config/mimeapps.list, so we do a little
+# trickery to change default applications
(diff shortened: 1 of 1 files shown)


renner0e:
this at least seems to work when building a with `just disk-image` and then installing firefox and thunderbird + logging out and back in
image

inffy (approved):

PR #2706, opened 2026-08-10
chore: revert kernel pin on stable

inffy:
https://github.com/ublue-os/akmods/pull/566 was merged so we are good to go with 7.1 kernel


renner0e:
needs to be put into main branch, stable should not have changes that are not in main

renner0e (approved):

PR #2734, opened 2026-08-18
feat: workaround for trimming physical root block device

renner0e:
https://github.com/fedora-silverblue/issue-tracker/issues/689

### DIFF
--- a/system_files/shared/usr/lib/systemd/system/fstrim.service.d/workaround-no-root-trim.conf
+++ b/system_files/shared/usr/lib/systemd/system/fstrim.service.d/workaround-no-root-trim.conf
@@ -3,4 +3,5 @@ Description=Discard unused blocks on mounted filesystems
 Documentation=man:fstrim(8) https://forge.fedoraproject.org/atomic/tracker/issues/81
 
 [Service]
+ExecStart=
 ExecStart=/usr/bin/fstrim --listed-in /proc/self/mountinfo --verbose --quiet-unsupported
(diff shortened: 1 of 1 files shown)


bsherman (commented):
Very interesting as a possible addition to `main`. I look forward to seeing how this goes.

renner0e:
so I tested this on a VM and it seems to work there

```
sudo /usr/bin/fstrim --listed-in /proc/self/mountinfo --verbose --quiet-unsupported
[sudo] password for user:
/boot/efi: 585.7 MiB (614133760 bytes) trimmed on /dev/sda1
/boot: 1.6 GiB (1770676224 bytes) trimmed on /dev/sda2
/etc: 35.2 GiB (37839376384 bytes) trimmed on /dev/sda3
```

On my host this doesn't work

`findmnt /etc/` seems to hint that for whatever reason my system doesn't have `discard=async`

host (installed 3 years ago from ostree fedora kinoite iso):

```
btrfs  rw,relatime,seclabel,compress=zstd:1,ssd,space_cache=v2,subvolid=258
```

vm (installed from our ISO):

```
btrfs  rw,relatime,seclabel,discard=async,space_cache=v2,subvolid=256,subvol=/
```

averms:
> so I tested this on a VM and it seems to work there

what's the output of `systemctl show fstrim.service -p ExecStart`? my guess is since it's a oneshot service it will show two commands as output: the original and the one you added. when the service is activated it will run both commands in serial.[^1]

the fix is to clear the ExecStart list with a
```
ExecStart=
```
before the desired fstrim command

[^1]: https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#ExecStart=

renner0e:
@averms yeah that was indeed the case, thanks so much for pointing this out, learned something new today about oneshot services I guess :slightly_smiling_face:

renner0e:
This is ready now

inffy (approved):

PR #2738, opened 2026-08-19
fix: install patched ostree to fix flatpak delta updates

renner0e:
This version from our copr has the fix for this issue applied.

https://github.com/ostreedev/ostree/issues/3635


renner0e:
we will want to make a stable release after we have verified that at least updates with bootc/rpm-ostree don't break with this applied, to verify if this patch worked, run flatpak update, I know that firefox was affected for me, it's the default browser here so I'm guessing most people use that, and it should be up-to-date as well!

inffy (approved):

PR #2768, opened 2026-08-24
fix: remove stale mariadb SELinux module from local policy store

### LINKED ISSUE #2767: Failure to upgrade due to mariadb-plugin-cracklib-password-check SELinux module
I maintain a customized version of Aurora that is mostly a slimming down of the dx image. I had a system that was failing to upgrade with the following message being thrown during the execution of `ostree-finalize-staged.service` on shutdown:

```
libsemanage.semanage_direct_get_module_info: Unable to open
mariadb-plugin-cracklib-password-check module lang ext file at
/etc/selinux/targeted/tmp/modules/200/mariadb-plugin-cracklib-password-check/lang_ext.
(No such file or directory).
```
This is the same issue that prompted #2026; the MariaDB installation that was re-added in this PR was then re-added to my custom image until I removed it in [hal7df/borealos#36](https://github.com/hal7df/borealos/issues/36). Most of my systems seemed to make it past this update without issue, except for this one, which was stuck on the last image version that had the MariaDB packages installed. I considered adding the `mariadb-cracklib-password-check` package back to my image to fix it, but ultimately worked around the issue by running the following as root:

```sh
semodule -X 200 -r mariadb-plugin-cracklib-password-check
semodule -B
bootc upgrade
```
However, it seems like it should be possible to at least switch/upgrade cleanly between images that have `mariadb-cracklib-password-check` installed and those that do not. I'm honestly not sure how to handle the Kinoite case, since the only "fix" that I can think of would be to add a systemd service that deletes the mariadb selinux module if it exists.

HuntedRaven7:
Fixes [issue 2767](https://github.com/ublue-os/aurora/issues/2767)

The mariadb-cracklib-password-check package (pulled in by the kinoite base via akonadi-server-mysql) installs the
mariadb-plugin-cracklib-password-check SELinux module into the persistent /etc/selinux policy store. Remnants of it break ostree-finalize-staged during upgrades or rebases onto images that do not ship the package, leaving systems unable to complete updates.

Add a oneshot systemd unit which removes the module from the local policy store at boot, including stale transaction leftovers, so that switching between images with and without the package finalizes cleanly.

### DIFF
--- a/build_scripts/base/17-cleanup.sh
+++ b/build_scripts/base/17-cleanup.sh
@@ -25,6 +25,10 @@ systemctl enable flatpak-nuke-fedora.service
 # DO NOT REMOVE THIS
 systemctl enable rechunker-group-fix.service
 
+# Remove stale mariadb SELinux module from the local policy store
+# https://github.com/ublue-os/aurora/issues/2767
+systemctl enable aurora-selinux-cleanup.service
+
 # run flatpak preinstall once at startup
 systemctl enable flatpak-preinstall.service
 
--- a/system_files/shared/usr/bin/aurora-selinux-cleanup
+++ b/system_files/shared/usr/bin/aurora-selinux-cleanup
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# Remove the mariadb-plugin-cracklib-password-check module from the local SELinux
+# policy store. The module gets installed into /etc/selinux by the
+# mariadb-cracklib-password-check package, and since /etc persists between
+# updates its remnants break `ostree-finalize-staged` (`semodule --refresh`)
+# when switching to images without the package:
+# https://github.com/ublue-os/aurora/issues/2767
+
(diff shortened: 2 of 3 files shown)


renner0e (changes requested):
I'm very hesitant to the very idea of this.

This needs to be reported upstream.

HuntedRaven7:
That is a fair point, I can go do that if need be.

renner0e:
> That is a fair point, I can go do that if need be.

That would be great. I'm not sure where to actually report this. Seems like ostree itself to me?

HuntedRaven7:
Uh no clue tbh it would make sense to do it there though.

HuntedRaven7:
https://github.com/ostreedev/ostree/issues/3563#issuecomment-5403144016 did a comment on a already made issue for this

HuntedRaven7:
Hey wanted to make an update! I made a pr over at https://github.com/ostreedev/ostree/pull/3649

PR #2774, opened 2026-08-25
feat: set Meta+Enter to open konsole

renner0e:
This is a very common shortcut and I've been using this for years. To my knowledge we can't make a standalone config for this so this will sadly have to be done this way.


ledif (approved):
I personally use Meta + Enter so LGTM 😆

Also, I didn't know about `X-KDE-Shortcuts`. Cool trick!

PR #2786, opened 2026-08-27
feat: switch to cardwire for managing multi-GPU systems

renner0e:
I'm not super familiar with these kinds of setups. This definitely needs a couple testers. Supposedly this is better than switcheroo.

Bazzite has a shim [1], I'm not sure how important this is for us as we use Flatpaked Gamerslop.

1: https://github.com/ublue-os/bazzite/blob/61a0554a8c5af47bb206884cdca25b64e51a9a9b/system_files/overrides/usr/bin/switcherooctl


inffy:
no hardware to test this at all but as far as I know this is the new thingy that has replaced switcheroo(ctl) which development has ceased (if i understood correctly).

renner0e:
Now that I think about it, I do might actually be able to test this, I also have an integrated GPU on my CPU

renner0e:
I think we may want to let this only be installed in testing for a while, I don't have particular reason other than being new.

Issue #2817, opened 2026-09-02
changelogs generally prefer non-rpm content and may misreport versions


renner0e:
Seen in the current release from today

The image should be on 7.1.8, but reports 7.1.12

https://github.com/ublue-os/aurora/releases/tag/stable-20260902.1

```
[root@480d356d65fe /]# rpm -qi kernel
Name        : kernel
Version     : 7.1.8
Release     : 200.fc44
```

renner0e:
This happened previously

https://github.com/ublue-os/aurora/issues/2324

inffy:
Not sure why this has happened just now, but the main reason seems to be that the kernel versions have diverged and how we "pin" the stable kernel.

Not really sure how the changelogs.py works in this matter, but looks like its picking up the upstream fedora kernel version now as its the first one.

renner0e:
So it seems like the presence of the `python3-perf` package is the reason it shows the old version, that is the only mention of "7.1.12" in our sbom. This is part of the kernel srpm.

This begs the question; is `python3-pef` something we should bundle in akmods perhaps so it stays in sync with whatever kernel we install?
CoreOS does not install it. Can we even easily do that?

There might also be a bug in some conditions in our images by shipping this mismatched package in general...

It seems like what we should probably do is *prefer*  `"foundBy": "rpm-db-cataloger"` over `"foundBy": "elf-binary-package-cataloger`.
If we would compile, let's say bootc from source in our images and don't at least remove bootc from the rpmdb we would *probably* run into the inverse issue. 

This is just a guess, but the first element in the list seems to be picked, which might've been the root cause for [this issue](https://github.com/ublue-os/aurora/issues/2077), the ostree objects just happened to be listed first here.

As a workaround we could also use the kernel-core package, which is reported correctly

<img width="645" height="162" alt="Image" src="https://github.com/user-attachments/assets/5c27f05e-f781-4525-98ed-2ae3720b50ea" />


```
# https://github.com/ublue-os/aurora/releases/tag/stable-20260902.1
oras pull ghcr.io/ublue-os/aurora:stable@sha256:c6d22acedcacc648c43040dcb56724eef44a5192bc7ca2672494119111469f88
```

<details>

```json
      "id": "f3300e5e48f92541",
      "name": "kernel",
      "version": "7.1.12-200.fc44",
      "type": "rpm",
      "foundBy": "elf-binary-package-cataloger",
      "locations": [
        {
[...]

PR #2818, opened 2026-09-02
fix(changelogs): use kernel-core instead of kernel

renner0e:
This happens to work around the fact that kernel in the SBOM collides with another package. It will now show the correct kernel, however, the underlying issue is not fixed. Besides that bug, we should probably be using this package anyway as kernel is just an empty metapackage with no actual content.

xref: https://github.com/ublue-os/aurora/issues/2817

### DIFF
--- a/.github/changelogs.py
+++ b/.github/changelogs.py
@@ -52,7 +52,7 @@ From previous `{target}` version `{prev}` there have been the following changes.
 ### Major packages
 | Name | Version |
 | --- | --- |
-| **Kernel** | {pkgrel:kernel} |
+| **Kernel** | {pkgrel:kernel-core} |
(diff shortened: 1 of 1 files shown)


renner0e:
local run of changelogs script

### Major packages
| Name | Version |
| --- | --- |
| **Kernel** | 7.1.6-201 ➡ 7.1.8-200 |
| **KDE** | 6.7.4-1 |
| **Mesa** | 26.1.4-4 |
| **Nvidia** | 610.57.04-1 |
| **Podman** | 5.8.4-1 |
| **Bootc** | 1.16.10-1 |
| **Flatpak** | 1.18.1-1 |
| **OSTree** | 2026.3-2 |
| **RPM-OSTree** | 2026.2-1 |

inffy (approved):

PR #2829, opened 2026-09-05
feat: make compatible with unsealed composefs

renner0e:
The bootc package ships a dracut configuration that adds the bootc
module (which is needed for the installation of the composefs backend)
to the initramfs which we currently don't have and ostree is also set
there.

I want this because bootc install is significantly faster when used with
the composefs backend, because it does some magic zeor-copy reflink
things. It is wired up with `just disk-image --backend composefs`.

For a long time systemd-boot was the only supported bootloader on
composefs backend, but grub works now. So we don't have to unnecessarily
add systemd-boot to the image.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -876,7 +876,7 @@ disk-image $image=default_image $tag=default_tag $flavor=default_flavor $ghcr="f
     if [[ "${backend}" == "ostree" ]]; then
       BOOTC_INSTALL_ARGS+=("--bootloader grub")
     else
-      BOOTC_INSTALL_ARGS+=("--bootloader systemd" "--composefs-backend")
+      BOOTC_INSTALL_ARGS+=("--bootloader grub" "--composefs-backend")
     fi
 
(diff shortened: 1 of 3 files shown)


renner0e:
for composefs backend install to-disk was failing with, no idea why that happens

```
Creating root filesystem (btrfs) on device /dev/loop0p3 (size=36.5 GB)
> mkfs.btrfs -U b1e95a75-20d2-4b78-aa0f-782d829e31cc -L root /dev/loop0p3
Creating ESP filesystem
> mkfs.fat /dev/loop0p2 -n EFI-SYSTEM
error: Installing to disk: Getting container info: failed to invoke method GetBlob: locating item named "sha256:ad38d10c0c9c4aa03422434a9555f98bd134722beb036def2de91b77fdfec4ca" for image with ID "6bbfd7ebae824e719bfa0ecdbff94dd67cc361fcabdb6c4ae742c499ea8b0269" (consider removing the image to resolve the issue): file does not exist
```

inffy (approved):

PR #2851, opened 2026-09-10
feat: vendor package repo configurations

renner0e:
We can get rid of the copr-helper wrapper script as we are no longer
enabling and then disabling the repo, which is explained in [1] as we
are configuring them to be off by default and then very purposefully
enabling it for that one dnf transaction where we actually need it
enabled.

The files in yum.repos.d were copied from my local /usr/etc directory.

[1] https://github.com/ublue-os/aurora/issues/1115


build_files/shared/etc/yum.repos.d/fedora-multimedia.repo
```diff
@@ -0,0 +1,35 @@
...
+gpgcheck=1
+enabled_metadata=1
+metadata_expire=6h
+type=rpm-md
+repo_gpgcheck=0
+
+[fedora-multimedia-source]
+name=negativo17 - Multimedia - Source
+baseurl=https://negativo17.org/repos/multimedia/fedora-$releasever/SRPMS
+enabled=0
```
renner0e, line 16: we could probably get rid of validate-repos.sh with this as well

build_scripts/shared/build-prep.sh
```diff
@@ -8,6 +8,15 @@ set -eoux pipefail
 cp /etc/dnf/dnf.conf /etc/dnf/dnf.conf.bak
 dnf config-manager setopt keepcache=1 timeout=60
 
+# configuring dnf repos
+rsync -rvKl /ctx/build_files/shared/ /
+
+if [[ "${IMAGE_FLAVOR}" == "dx" ]]; then
+  rsync -rvKl /ctx/build_files/dx/ /
+fi
```
renner0e, line 18: might be worth to make this one rsync call

build_files/shared/etc/yum.repos.d/fedora-multimedia.repo
renner0e: Might also be worth looking into allowlisting specific packages per repo...

renner0e: this PR is already big enough, although this is definitely something worth looking into as that could make it possible to make less but bigger dnf calls while still ensuring we don't install packages where we shouldn't get them from.

renner0e:
this is now ready for testing/review, this is very much a building block PR, this could allow the use case mentioned in https://github.com/ublue-os/aurora/pull/2851#discussion_r3990410252 and also allow vendoring the GPG keys of our 3rd party repos, currently they are fetched from the network, if they get compromised we would happily install their packages anyway.

inffy (approved):
looks fine to me, doesn't break anything locally

PR #2871, opened 2026-09-14
feat: add iwd notifier

### LINKED ISSUE #262: iwd is dead
Intel has killed IWD development. We still have the ujust in our scripts.

What should we do about it. Plainly removing it would be pretty much a no-go as there might be some users that have switched to it and then they would have no easy way back.

Think bazzite did some kind service or script that will move users back to wpa_supplicant, not sure how well it has worked and its another failure point.

inffy:
Adds a notifier (mostly copied the zfs one) which shows a notification weekly to the users that might have enabled iwd with ujust.

Needs: https://github.com/get-aurora-dev/common/pull/276

Closes: https://github.com/get-aurora-dev/common/issues/262


renner0e (commented):
hmm I'm getting this:

```
sudo systemctl start --verbose aurora-iwd-deprecation-notifier.service
```

```
Deprecation Notifier...
machinectl[80397]: Connected to the local host. Press ^] three times within 1s to exit session.
aurora-iwd-deprecation-notifier[80397]: Invalid number of options.
machinectl[80397]: Connection to the local host terminated.
systemd[1]: aurora-iwd-deprecation-notifier.service: Deactivated successfully.
systemd[1]: Finished aurora-iwd-deprecation-notifier.service - Aurora IWD

```

PR #2874, opened 2026-09-15
chore: pin amd-gpu-firmware because of regressions

inffy:
relates to https://gitlab.freedesktop.org/drm/amd/-/work_items/5803


build_scripts/base/01-packages.sh
```diff
@@ -173,6 +173,10 @@ dnf -y remove "${EXCLUDED_PACKAGES[@]}"
 #    dnf5 upgrade --refresh --advisory=FEDORA-2024-dd2e9fb225
 #fi
 
+# AMD GPU firmware has regressions for 680M igpus
+# https://gitlab.freedesktop.org/drm/amd/-/work_items/5803
+dnf -y swap amd-gpu-firmware amd-gpu-firmware-20260810-1.fc44
```
renner0e: We should always use, going from memory here sorry

rpm -E %fedora, instead of hardcoding the fedora version

renner0e:
Applied the Backport label, I think this is important enough to warrant a new stable image

inffy:
> Applied the Backport label, I think this is important enough to warrant a new stable image

i think the backport action is broken at the moment :/

renner0e:
Manual is fine of course, I see that this downgrades all the other firmware packages as well, shouldn't cause any issues to my knowledge

image-versions.yml
```diff
@@ -9,7 +9,7 @@ images:
   - name: common
     image: ghcr.io/get-aurora-dev/common
     tag: latest
-    digest: sha256:1f301431774cd7a432886214aa3955a4cdec6f36e927a45f7e39984cc446c0d6
+    digest: sha256:ea65f0835a2c45f644bd76f679330e2734fa265458302a0570472581d578f208
```
renner0e: Intended?

renner0e (approved):

PR #2877, opened 2026-09-15
fix(backport-action): keep credentials so backport push works

inffy:
the backport action needs the credentials to push the backport PRs


renner0e (approved):
Thanks

PR #2884, opened 2026-09-16
chore(ci): move to newer free-disk-space action when needed

### LINKED ISSUE #2864: remove remove-unwanted-software-action
- we should find out how much space we need on the biggest configuration (should be the normal image due to bootc install, otherwise it should be normal aurora-dx)
- switch over to https://github.com/HastD/free-disk-space

Seems like github runners have consistent disk configurations now, we would save 30s-2 min per run with this.

inffy:
Looked at few of the recent PR runs and it looks like the runners seem to have pretty much consistently ~92gigs of space.

That should be enough for any of our images pretty much any time

Closes #2864


inffy:
This should be fine. We should be able to drop the "minumum" to something like 65 if we want. But it looks like the default free space now is consistently 92G which should be way over what our biggest builds need

renner0e (approved):
Let's hope GitHub doesn't change their minds lol

PR #2887, opened 2026-09-17
feat: bump testing to Fedora 45

renner0e:
see commits

this is supposed to get the builds going only, there are some things to cleanup/implement which would be nice to have which we can do later


renner0e:
fresh image boots and containers signing policy is fine, if the CI is green here, this is fine to merge then

inffy (approved):

